PR #25, opened 2021-05-14
Psr7 Uploaded File: use stateless properties to cache already done moves

Slamdunk:
Hi, I'm the author of the original PR that allowed filtering multiple times: https://github.com/zendframework/zendframework/pull/3749

I'm using PSR7 now, and I am unable to call both `$form->isValid();` and `$form->getData();` consequentially, because the filter attempts to call `\Psr\Http\Message\UploadedFileInterface::getStream` twice:

https://github.com/laminas/laminas-filter/blob/76a81282404795302c188ed25a5b7afcc8b7a889/src/File/RenameUpload.php#L389-L395

This is an issue since `\Psr\Http\Message\UploadedFileInterface` may be stateful, like the Diactoros one: you can only call `getStream` before the `moveTo`; after that an exception is raised:

https://github.com/laminas/laminas-diactoros/blob/bcb2daf4201fb2a1090522e42b163429c5be7b99/src/UploadedFile.php#L139-L141

So the `uri` cannot be used as a key for the internal cache.

`spl_object_hash` is good though.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.6.1@e93e532e4eaad6d68c4d7b606853800eaceccc72">
+<files psalm-version="4.6.3@f1a840727dd756899eee2f1f9ea443e265a4763f">
   <file src="src/AbstractDateDropdown.php">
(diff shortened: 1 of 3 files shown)


Slamdunk:
Ping @boesing or @weierophinney

weierophinney (approved):

PR #29, opened 2021-07-13
Filter for arithmetic operations #28

### LINKED ISSUE #28: Filter for arithmetic operations
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

I created this filter for an application that converts values from APIs. It's useful for measure unit conversions and making adjustments. For example, an endpoint returns the amount of RAM requested by an user and you can add a security margin.

fgsl:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

I created this filter for an application that converts values from APIs. It's useful for measure unit conversions and making adjustments. For example, an endpoint returns the amount of RAM requested by an user and you can add a security margin.


boesing:
No worries, there is no way to let you fix the codestyle.
Ignore it for now, its a known issue with running coding-standard v1 on PHP 7.4.

fgsl:
> No worries, there is no way to let you fix the codestyle.
> Ignore it for now, its a known issue with running coding-standard v1 on PHP 7.4.

Thank you and excuse me. I thought I must format the code again and sent a new change...

src/FourOperations.php
```diff
@@ -0,0 +1,63 @@
...
+            self::MUL,
+            self::DIV,
+            self::MOD
+        ];
+        if (! isset($options['operation']) || (! in_array($options['operation'], $operations))) {
+            throw new InvalidArgumentException(sprintf('%s expects argument operation string with one of theses values: add, sub, mul or div; received "%s"', __METHOD__, (float) $options['operation']));
+        }
+        if (! isset($options['value'])) {
+            throw new InvalidArgumentException(sprintf('%s expects argument value; received none', __METHOD__));
+        }
```
froschdesign: Please allow a creation of the filter without options, this would be consistent with other filters and their behaviour in the current version. Setter and getters are needed then, also the related unit tests.
Compare with the `PrepReplace` filter:

https://github.com/laminas/laminas-filter/blob/2c5e0f440bfd0de8e4c09e28269249a8785f1185/src/PregReplace.php#L121-L135

src/FourOperations.php
```diff
@@ -0,0 +1,63 @@
+<?php
+namespace Laminas\Filter;
+
+use Laminas\Filter\AbstractFilter;
+use Laminas\Filter\Exception\InvalidArgumentException;
+
+class FourOperations extends AbstractFilter
```
froschdesign: The _four_ operations are addition, subtraction, multiplication and division but here we have one more.

test/FourOperationsTest.php
```diff
@@ -0,0 +1,38 @@
+<?php
+namespace LaminasTest\Filter;
+
+use Laminas\Filter\FourOperations;
+use PHPUnit\Framework\TestCase;
+
+class FourOperationsTest extends TestCase
+{
+
+    public function testOperations(): void
```
froschdesign: A [data provider](https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html#data-providers) can be used here.

test/FourOperationsTest.php
```diff
@@ -0,0 +1,38 @@
+<?php
+namespace LaminasTest\Filter;
+
+use Laminas\Filter\FourOperations;
+use PHPUnit\Framework\TestCase;
+
+class FourOperationsTest extends TestCase
```
froschdesign: Please test also invalid data types.

froschdesign (changes requested):
We definitely need some documentation that shows real use cases for this filter.

src/FourOperations.php
fgsl: Done!

src/FourOperations.php
fgsl: Done!

test/FourOperationsTest.php
fgsl: Done!

test/FourOperationsTest.php
fgsl: Done!

fgsl:
I made a little change on typing, but I don't know I had to follow all the messages in https://github.com/laminas/laminas-filter/actions/runs/1033956242. I think some parameters must be mixed, because they can be integers or floats.

src/FiveOperations.php
```diff
@@ -0,0 +1,115 @@
+<?php
+namespace Laminas\Filter;
+
+use Laminas\Filter\AbstractFilter;
+use Laminas\Filter\Exception\InvalidArgumentException;
+
+class FiveOperations extends AbstractFilter
```
Ocramius: IMO worth having separate filters, than one named "five operations" that merges them all together: having one thing that does multiple separate things is at the root of most accidental complexity in software.

Ocramius (changes requested):

fgsl:
Hi, @Ocramius. I have splitted the new filter into five filters, according your guideline.

src/Add.php
```diff
@@ -0,0 +1,24 @@
...
+     * @see \Laminas\Filter\FilterInterface::filter()
+     */
+    public function filter($value)
+    {
+        if (! isset($this->options['operand'])) {
+            throw new InvalidArgumentException(sprintf('%s expects argument value; received none', __METHOD__));
+        }
+
+        $value = (float) $value;
+        $operand = (float) $this->options['operand'];
```
weierophinney: Why the cast to float here?

If I were operating on integers, I'd expect an integer return value. Perhaps this should be broken into "AddInteger" and "AddFloat"?

src/Sub.php
```diff
@@ -0,0 +1,24 @@
...
+     * @see \Laminas\Filter\FilterInterface::filter()
+     */
+    public function filter($value)
+    {
+        if (! isset($this->options['operand'])) {
+            throw new InvalidArgumentException(sprintf('%s expects argument value; received none', __METHOD__));
+        }
+
+        $value = (float) $value;
+        $operand = (float) $this->options['operand'];
```
weierophinney: Same remark as for the `Add` filter; should this maybe be broken into `SubInt` and `SubFloat`?

src/Mod.php
```diff
@@ -0,0 +1,24 @@
...
+     */
+    public function filter($value)
+    {
+        if (! isset($this->options['operand'])) {
+            throw new InvalidArgumentException(sprintf('%s expects argument value; received none', __METHOD__));
+        }
+
+        $value = (float) $value;
+        $operand = (float) $this->options['operand'];
+        return ($value % $operand);
```
weierophinney, line 22: This doesn't do what you think it does. From the manual:

> Operands of modulo are converted to int before processing. For floating-point modulo, see fmod().

In other words, using the `%` operator will always return an integer. If you want to work with floats, you need to use `fmod()`. So, perhaps two variants, `Mod` and `FloatMod`, are needed.

src/Mul.php
```diff
@@ -0,0 +1,24 @@
...
+     */
+    public function filter($value)
+    {
+        if (! isset($this->options['operand'])) {
+            throw new InvalidArgumentException(sprintf('%s expects argument value; received none', __METHOD__));
+        }
+
+        $value = (float) $value;
+        $operand = (float) $this->options['operand'];
+        return ($value * $operand);
```
weierophinney: Same remark as for `Add` and `Sub`: if integers are used, I'd expect an integer return value, not casting to float.

weierophinney (changes requested):
First, this needs to be rebased against the current release branch (currently 2.13.x, but likely to become 2.14.x later today when the PHP 8.1 support is merged); there are a few artifacts in here that seem to be from already merged changes (specifically the `AllowList` and `DenyList` implementations and tests). 

Second, I've noted a number of areas where the behavior of the filters would not behave as expected when using integer values and operands. These likely need to be split into additional filters to allow both integer and float operations (with the exception of the `Div` filter, where a float is always expected).

gsteel:
As this patch is getting on for 4 years old, and looks abandoned - I'm closing for now. Feel free to send in another patch, but please bear in mind that v3 is close to release with some significant changes.

PR #31, opened 2021-10-23
Use inclusive language: deprecate `BlackList` and `WhiteList` filters in favour of `DenyList` and `AllowList`

becoded:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Use inclusive language for the filters


src/Blacklist.php
```diff
@@ -2,82 +2,22 @@
...
-
-    /**
-     * {@inheritDoc}
-     *
-     * Will return null if $value is present in the black-list. If $value is NOT present then it will return $value.
-     */
-    public function filter($value)
-    {
-        return in_array($value, $this->getList(), $this->getStrict()) ? null : $value;
+        trigger_error(
```
Ocramius: IMO no loud deprecation to be used: they've been a massive PITA for everyone, and should be avoided overall.

becoded: Fair point, changed it.

Ocramius (approved):
LGTM, thanks @becoded

PR #35, opened 2021-11-28
Added support for PHP 8.1

### LINKED ISSUE #34: PHP 8.1 support
I was wondering if PHP 8.1 support is planned for laminas-filter. At the moment I'm getting a lot of deprecated warnings:
```
PHP Deprecated:  Return type of Laminas\InputFilter\BaseInputFilter::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in vendor\laminas\laminas-inputfilter\src\BaseInputFilter.php on line 70
PHP Deprecated:  Return type of Laminas\Filter\FilterChain::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in vendor\laminas\laminas-filter\src\FilterChain.php on line 92
```

driehle:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | yes
| QA            | no

### Description

This PR adds support for PHP 8.1. Besides, laminas-coding-standard has been updated to 2.0.

Note that I have suppressed a couple of type hint errors in `phpcs.xml`. Those rules are simply too noisy and I cannot fix all the issues there.


phpcs.xml
```diff
@@ -1,8 +1,18 @@
...
+
+    <!-- Ignore warnings, show progress of the run and show sniff names -->
+    <arg value="nps"/>
 
     <!-- Paths to check -->
     <file>src</file>
     <file>test</file>
+
+    <exclude-pattern>src/Int.php</exclude-pattern>
+    <exclude-pattern>src/Null.php</exclude-pattern>
```
tobias-trozowski, line 22: we should follow the same structure here as in other components. 

@see https://github.com/laminas/repo-template/blob/main/phpcs.xml.dist

src/Compress/Gz.php
```diff
@@ -152,19 +173,19 @@ public function compress($content)
...
+     * @throws Exception\RuntimeException If unable to open archive or error during decompression.
      */
     public function decompress($content)
     {
         $archive = $this->getArchive();
         $mode    = $this->getMode();
 
         //check if there are null byte characters before doing a file_exists check
-        if (false === strpos($content, "\0") && file_exists($content)) {
+        if (null !== $content && false === strpos($content, "\0") && file_exists($content)) {
```
tobias-trozowski, line 184: as unchanged lines cannot be commented i throw it here...
given `$content` is `null` it would raise an error further below:
```
} elseif ($mode === 'deflate') {
    $compressed = gzinflate($content);
} else {
    $compressed = gzuncompress($content);
}
```

this seems to be uncovered in tests (`DecompressTest` is testing this with bz2)

src/DenyList.php
```diff
@@ -40,7 +41,7 @@ public function setStrict($strict = true): void
     /**
      * Returns whether the in_array() call should be "strict" or not. See in_array docs.
      *
-     * @return boolean
+     * @return Boolean
```
tobias-trozowski:
```suggestion
     * @return bool
```

phpcs.xml
driehle: Not sure what you mean, the linked file doesn't contain any exclude patterns.

tobias-trozowski: sorry, my fault. i meant the overall structure of it.
e.g.
```xml
<?xml version="1.0"?>
<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="./vendor/squizlabs/php_codesniffer/phpcs.xsd">

    <arg name="basepath" value="."/>
    <arg name="cache" value=".phpcs-cache"/>
    <arg name="colors"/>
    <arg name="extensions" value="php"/>
    <arg name="parallel" value="80"/>

    <!-- Show progress -->
    <arg value="p"/>

    <!-- Paths to check -->
    <file>src</file>
    <file>test</file>

    <exclude-pattern>src/Int.php</exclude-pattern>
    <exclude-pattern>src/Null.php</exclude-pattern>

    <!-- Include all rules from Laminas Coding Standard -->
    <rule ref="LaminasCodingStandard">
        <exclude name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingAnyTypeHint"/>
        <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingAnyTypeHint"/>
        <exclude name="SlevomatCodingStandard.TypeHints.PropertyTypeHint.MissingAnyTypeHint"/>
    </rule>
</ruleset>
```

driehle: Fixed with 448ac52

src/DenyList.php
driehle: Fixed with 448ac52

src/Compress/Gz.php
driehle: Indeed, I added an explicit check for `null` values and extended tests in 73cd4bd

src/FilterPluginManager.php
```diff
@@ -10,6 +12,29 @@
...
+use zend\filter\word\dashtocamelcase;
+use zend\filter\word\dashtoseparator;
+use zend\filter\word\dashtounderscore;
+use zend\filter\word\separatortocamelcase;
+use zend\filter\word\separatortodash;
+use zend\filter\word\separatortoseparator;
+use zend\filter\word\underscoretocamelcase;
+use zend\filter\word\underscoretodash;
+use zend\filter\word\underscoretoseparator;
+use zend\filter\word\underscoretostudlycase;
```
driehle: It looks like PHPCs has broken the imports to the old Zend classes. Any ideas how to solve this?

test/DecompressTest.php
```diff
@@ -73,56 +101,73 @@ public function testCompressToFile()
...
-        return [
-            [null],
-            [new \stdClass()],
-            [[
-                'decompress me',
-                'decompress me too, please'
-            ]]
-        ];
+        foreach ($this->returnFilterType() as $parameter) {
+            yield [$parameter[0], new stdClass()];
```
tobias-trozowski, line 140: previously available `null` was missing here
```suggestion
            yield [null];
            yield [$parameter[0], new stdClass()];
```

driehle: It went further down into `testDecompressNullValueThrowsRuntimeException`. As that one is supposed to throw an exception, handling it in the test case above doesn't make sense to me.

src/FilterPluginManager.php
driehle: Restored aliases in da43764 and put them in a `@codingStandardsIgnoreStart/End` section, to prevent them from being changed by PHPCs again.

test/DecompressTest.php
tobias-trozowski: in the previous version the test was expecting `null` as return when `null` was provided as argument, so this PR is actually introducing a new behaviour.

driehle: Ok, I see. It looks like `bzdecompress(null)` actually works:

```
driehle:/laminas-filter$ php7.3 -r "var_dump(bzdecompress(null));"
string(0) ""
driehle:/laminas-filter$ php7.4 -r "var_dump(bzdecompress(null));"
string(0) ""
driehle:/laminas-filter$ php8.0 -r "var_dump(bzdecompress(null));"
string(0) ""
driehle:/laminas-filter$ php8.1 -r "var_dump(bzdecompress(null));"
string(0) ""
```

That is different from `gzuncompress(null)`, which has never worked and which throws a type incompatibility error with PHP 8.1:

https://3v4l.org/VJfrk

```
Deprecated: gzuncompress(): Passing null to parameter #1 ($data) of type string is deprecated in /in/VJfrk on line 3

Warning: gzuncompress(): data error in /in/VJfrk on line 3
bool(false)

Fatal error: Uncaught Error: Call to undefined function bzdecompress() in /in/VJfrk:6
Stack trace:
#0 {main}
  thrown in /in/VJfrk on line 6
```

So we would have to keep accepting `decompress(null)` for the `Bz2` filter, but allow it with the `Gz` filter... :-\

tobias-trozowski: i would love to see all this fully typed and non-null... but yeah... need to keep accepting this.

driehle: Well, there is an issue. Updating to laminas-coding-standard 2.0 has introduced `declare(strict_types=1)` in all files. This doesn't allow passing `null` to `bzuncompress()` anymore withouth getting a `TypeError`. I didn't recognize this at first, as my console commands posted above didn't include strict types. Here is the updated output:

```
driehle:/laminas-filter$ php7.3 -r "declare(strict_types=1);bzdecompress(null);"
PHP Fatal error:  Uncaught TypeError: bzdecompress() expects parameter 1 to be string, null given in Command line code:1
Stack trace:
#0 Command line code(1): bzdecompress(NULL)
#1 {main}
  thrown in Command line code on line 1
```

Output is the same for PHP 7.3, 7.4, 8.0 and 8.1.

I'll add something to cover null explicitly in the code.

tobias-trozowski: the declare strict only kicks in for files which define it. if a have a `test.php` w/o `declare(strict_types=1)` there wont be a type error.

https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict

> Strict typing applies to function calls made from within the file with strict typing enabled, not to the functions declared within that file. If a file without strict typing enabled makes a call to a function that was defined in a file with strict typing, the caller's preference (coercive typing) will be respected, and the value will be coerced.

driehle: You are right, but strict types are declared in our tests as well. I still think that we should cover this in the source files, so that it works for both users having strict types enabled or not enabled. See 289743d for a proposed solution.

src/Compress/Gz.php
driehle: See https://github.com/laminas/laminas-filter/pull/35#pullrequestreview-817519564 for follow up discussion on null values.

test/Compress/GzTest.php
```diff
@@ -177,4 +186,15 @@ public function testGzToString()
...
+
+    public function testGzDecompressNullThrowsRuntimeException()
+    {
+        $this->expectException(Exception\RuntimeException::class);
+        $this->expectExceptionMessage('Error during decompression');
+
+        $filter = new GzCompression();
+        $result = $filter->decompress(null);
+
+        $this->assertEmpty($result);
```
tobias-trozowski: shouldn't be needed as decompress method is throwing exception with null arg

driehle: Fixed with b9e3ef4

driehle:
Is there something I can further help with?

PR #39, opened 2021-11-29
Adds call-outs for new and deprecated features of version 2.12.0

froschdesign:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Enhancement        | yes

If a feature is marked as deprecated, it does not mean that we can remove it from the documentation without notice.
And new features also need to be highlighted.

Related to #31


Ocramius (approved):

Ocramius:
Thanks @froschdesign!

PR #40, opened 2021-12-03
Removed filter 'Null' and 'Int'

tobias-trozowski:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description
- removed deprecated filter 'Null' and 'Int' (both deprecated since 2015)


Ocramius (approved):
LGTM! 

Note: the reason why this can be merged in a **MINOR** release, is that `Int` and `Null` are not instantiable as per PHP 7.3.0+, and therefore they can be safely removed, without any side-effect for consumers (they would get a crash when attempting to instantiate the currently defined symbols anyway).

For example ( https://3v4l.org/sH2sr )

```php
<?php

namespace Laminas\Filter;

class Int
{
}

var_dump(new Int());
```

Produces:

```
Fatal error: Cannot use 'Int' as class name as it is reserved in /in/sH2sr on line 5

Process exited with code 255.
```

Issue #41, opened 2021-12-13
Non-string scalar types produce a fatal error with StripNewlines


internalsystemerror:
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.13.0

#### Summary

On PHP 7.4, `str_replace` would accept any scalar value and implicitly cast it to a string (even with `declare(strict_types=1)`). As of PHP 8.0, anything but `string|array` to the 3rd argument of `str_replace` throws a fatal error.

See: https://3v4l.org/YOktG

With the PHP 8.1 release, `declare(strict_types=1)` was added to all files (presumably due to the coding standard) meaning that previously valid values passed to `StripNewLines` are now throwing a fatal error.

#### Current behavior

`StripNewLines::filter(123)` throws a fatal error on the latest version whereas previously, it would be implicitly cast to a string.

#### How to reproduce

I intend to submit a PR to introduce failing tests along with a fix.

#### Expected behavior

Previously working values should not throw a fatal error.

Ocramius:
The declaration says that it accepts `string|array` though, not `mixed` :thinking_face:

https://github.com/laminas/laminas-filter/blob/29db2ddc0671d7a250454a466f56922d12d4b255/src/StripNewlines.php#L18-L19

I think we should adjust it to allow `mixed`, then filter only what is filterable?

From `AbstractFilter`, I can see that this is completely broken/wrong:

https://github.com/laminas/laminas-filter/blob/29db2ddc0671d7a250454a466f56922d12d4b255/src/AbstractFilter.php#L84-L96

internalsystemerror:
> The declaration says that it accepts `string|array` though, not `mixed` :thinking_face:

> I think we should adjust it to allow `mixed`, then filter only what is filterable?

This is approach taken in the PR, to expect `$value` to be mixed.

Ocramius:
Linking the PR, since I hadn't seen it before :D 

https://github.com/laminas/laminas-filter/pull/42

internalsystemerror:
@Ocramius this is because you are a very busy person and your time is precious. This issue is mentioned in the description of the PR, so they should already be linked but thank you.

Ocramius:
Sitting in boring meetings, so it isn't too bad :P

Ocramius:
Handled in #42

PR #42, opened 2021-12-13
Fix fatal errors due to strict types in string filters

internalsystemerror:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

As per issue https://github.com/laminas/laminas-filter/issues/41, this PR ensures that non-string scalar types (int|float|bool) still pass through the filters as before. The classes I identified with the problem are:
- `PregReplace::class`
- `StripNewlines::class`
- `Word/CamelCaseToSeparator::class`
- `Word/DashToSeparator::class`
- `Word/SeparatorToCamelCase::class`
- `Word/SeparatorToSeparator::class`
- `Word/UnderscoreToStudlyCase::class`

In this PR are 2 commits. The first ensures thats all tests are checking for valid types now that `strict_types` is used. The second adds the fixes described above.

A couple of extra things that should be pointed out. The BooleanFilter was previously expecting null to be filtered to null, with all types and no casting. This is incorrect as TYPES_ALL includes TYPE_NULL. Changing to use `assertSame` instead of `assertEquals` caught this error.

There is also a CS ignore for a deprecated OpenSSL method to allow the CS checks to pass as there is no alternative method, as the functionality it provided is now default.

### DIFF
--- a/src/AbstractFilter.php
+++ b/src/AbstractFilter.php
@@ -107,6 +107,8 @@ abstract class AbstractFilter implements FilterInterface
     }
 
     /**
+     * @internal
+     *
      * @param  mixed $value
      * @return mixed
      */
--- a/src/AbstractFilter.php
+++ b/src/AbstractFilter.php
@@ -8,10 +8,12 @@ use Laminas\Stdlib\StringUtils;
 use Traversable;
 
 use function array_key_exists;
+use function array_map;
 use function get_class;
 use function gettype;
 use function is_array;
 use function is_object;
+use function is_scalar;
 use function method_exists;
 use function sprintf;
 use function str_replace;
@@ -102,5 +104,20 @@ abstract class AbstractFilter implements FilterInterface
     protected static function isOptions($options)
     {
         return is_array($options) || $options instanceof Traversable;
+    }
+
+    /**
+     * @param  mixed $value
+     * @return mixed
+     */
+    protected static function applyFilterOnlyToStringableValuesAndStringableArrayValues($value, callable $callback)
+    {
+        if (! is_array($value)) {
+            if (! is_scalar($value)) {
+                return $value;
+            }
+            return $callback((string) $value);
+        }
+        return $callback(array_map(fn($item) => is_scalar($item) ? (string) $item : $item, $value));
     }
 }
--- a/src/PregReplace.php
+++ b/src/PregReplace.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Laminas\Filter;
 
+use Closure;
 use Traversable;
 
 use function func_get_args;
@@ -11,7 +12,6 @@ use function get_class;
 use function gettype;
 use function is_array;
 use function is_object;
-use function is_scalar;
 use function is_string;
 use function iterator_to_array;
 use function preg_match;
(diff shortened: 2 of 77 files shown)


src/PregReplace.php
```diff
@@ -137,8 +137,13 @@ public function getReplacement()
      */
     public function filter($value)
     {
-        if (! is_scalar($value) && ! is_array($value)) {
-            return $value;
+        if (! is_array($value)) {
```
Ocramius: What happens if the array contains non-scalars?

src/StripNewlines.php
```diff
@@ -15,13 +16,18 @@ class StripNewlines extends AbstractFilter
...
-     * @param  string|array $value
-     * @return string|array
+     * @param  mixed $value
+     * @return mixed
      */
     public function filter($value)
     {
-        if (! is_scalar($value) && ! is_array($value)) {
-            return $value;
+        if (! is_array($value)) {
```
Ocramius: Same question as above: what happens to non-scalars? Perhaps we need a small `@internal` utility class that does this `is_array()`, `is_scalar()` and `is_string()` check recursively, somehow? :thinking:

Ocramius (changes requested):

src/StripNewlines.php
Ocramius: name not important, but literally a common function (or static method) such as `apply_filter_only_to_stringable_and_stringable_array_values(mixed $value, callback $filter)`

internalsystemerror: Do we want to do the same implicit casting to the array values?

If one of the array values contains an object, should the whole array be returned untouched? My only apprehension to this is BC. When considering how str_replace would have handled this previously:

Consider the scenario, you have a custom filter to change a collection of userIds into a collection of `User` objects, and that User class implements `__toString()`. Would `str_replace` cast these objects to a string implicitly previously?

This is a fairly contrived example I admit, but it demonstrates my thoughts regarding changing existing behaviour. This is starting to give me a headache :D.

src/PregReplace.php
internalsystemerror: I will update this according to the solution decided below.

src/StripNewlines.php
Ocramius: The reason why I asked about `is_array()` and recursion is that I don't know if that's what the `preg_replace()` functions do...

internalsystemerror:
I've updated this PR with a 3rd commit to address the requested changes. The specific portion is the introduction of a new static method on `AbstractFilter`:

```php
protected static function applyFilterOnlyToStringableValuesAndStringableArrayValues($value, callable $callback)
{
    if (! is_array($value)) {
        if (! is_scalar($value)) {
            return $value;
        }
        return $callback((string) $value);
    }
    return $callback(array_map(fn($item) => is_scalar($item) ? (string) $item : $item, $value));
}
```

This should replicate the functionality as it was prior to adding `declare(strict_types=1)`. If it would error previously, it should error now, and if it implicitly cast previously, it should implicitly cast now.

As for multidimensional arrays, my guess from reading php.net is that these are not handled by `preg_replace` according to https://www.php.net/manual/en/function.preg-replace.php:

> replacement: The string or an array with strings to replace.

And although not specifically stated, I get the same impression from reading https://www.php.net/manual/en/function.str-replace:

> If search and replace are arrays, then str_replace() takes a value from each array and uses them to search and replace on subject.

EDIT: Using https://3v4l.org/nKlWk, a multidimensional array isn't supported, however it gives a Notice on PHP < 8..0 and a Warning on PHP >= 8.0.

src/AbstractFilter.php
```diff
@@ -103,4 +105,19 @@ protected static function isOptions($options)
     {
         return is_array($options) || $options instanceof Traversable;
     }
+
+    /**
+     * @param  mixed $value
+     * @return mixed
```
Ocramius, line 113: Let's make this `@internal`

Ocramius (changes requested):
One last change to go :D 

Just want to prevent abuse of the newly introduced utility...

src/AbstractFilter.php
internalsystemerror: tx! will do

Ocramius (approved):
:ship: 

Thanks @internalsystemerror!

PR #47, opened 2022-02-07
Avoid `null` return value from `Uri#getPath()` leading to a crash in `UriNormalize`

Bontah:
Signed-off-by: Dennis Plettner <d.plettner@adcell.de>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Using the UriNormalizer with enforcedScheme and an empty string will cause strpos() to fail because it expects a string.
uri->getPath() however can return a string and null. Now if the filter() function is called with an empty string then an uri object will be created with path being null.

i also provided a testcase for this specific situation.

### DIFF
--- a/src/UriNormalize.php
+++ b/src/UriNormalize.php
@@ -123,7 +123,8 @@ class UriNormalize extends AbstractFilter
      */
     protected function enforceScheme(Uri $uri)
     {
-        $path = $uri->getPath();
+        $path = $uri->getPath() ?? '';
+
         if (strpos($path, '/') !== false) {
             [$host, $path] = explode('/', $path, 2);
             $path          = '/' . $path;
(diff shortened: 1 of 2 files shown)


Ocramius (approved):
Besides the CS issue (please fix it up, if possible), LGTM

Bontah:
okay i think i got it now. 
quick question: would amand or a new commit be the right way in cases like this? i use gerrit and rebase/amand for work related stuff without pull requests so i don't know it exactly any better.

Ocramius:
> would amand or a new commit be the right way in cases like this? 

Anything that is mostly "cleanup" can generally be amended :)

Ocramius:
I re-targeted `2.13.x` - this is a bugfix.

Will rebase before merging.

Ocramius:
Rebased/squashed, thanks @Bontah!

PR #52, opened 2022-07-12
Improve type inference for the filter plugin manager

### LINKED ISSUE #50: Removal of `container-interop/container-interop`
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | yes

#### Summary

This issue is created to keep track of the status of getting rid `container-interop` in the `laminas` components.

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

- Updates `laminas-servicemanager` to 3.14, removing `interop-container`
- Adds template types to the plugin manager and filter chain
- Amends `StaticFilter` so that it correctly deals with possible callables retrieved from the plugin manager
- Replace references to Interop\Container with Psr\Container
- Update psalm configuration replacing deprecated attribute and enabling unused suppressions

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,20 +4,20 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "6bffd6715b62da088c9435e41795f7f8",
+    "content-hash": "b0857dedd40c2327a0ba75c65548ad1a",
     "packages": [
         {
             "name": "laminas/laminas-stdlib",
-            "version": "3.6.1",
+            "version": "3.10.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/laminas/laminas-stdlib.git",
-                "reference": "db581851a092246ad99e12d4fddf105184924c71"
+                "reference": "0d669074845fc80a99add0f64025192f143ef836"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/laminas/laminas-stdlib/zipball/db581851a092246ad99e12d4fddf105184924c71",
-                "reference": "db581851a092246ad99e12d4fddf105184924c71",
+                "url": "https://api.github.com/repos/laminas/laminas-stdlib/zipball/0d669074845fc80a99add0f64025192f143ef836",
+                "reference": "0d669074845fc80a99add0f64025192f143ef836",
                 "shasum": ""
             },
             "require": {
@@ -28,8 +28,8 @@
             },
             "require-dev": {
                 "laminas/laminas-coding-standard": "~2.3.0",
-                "phpbench/phpbench": "^0.17.1",
-                "phpunit/phpunit": "~9.3.7",
+                "phpbench/phpbench": "^1.0",
+                "phpunit/phpunit": "^9.3.7",
                 "psalm/plugin-phpunit": "^0.16.0",
                 "vimeo/psalm": "^4.7"
             },
@@ -63,7 +63,7 @@
                     "type": "community_bridge"
                 }
             ],
-            "time": "2021-11-10T11:33:52+00:00"
+            "time": "2022-06-10T14:49:09+00:00"
         }
     ],
     "packages-dev": [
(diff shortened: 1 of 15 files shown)


composer.lock
```diff
@@ -451,42 +451,6 @@
             ],
             "time": "2021-07-31T17:03:58+00:00"
         },
-        {
-            "name": "container-interop/container-interop",
```
Ocramius, line 455: It is gone :partying_face:

src/FilterChain.php
```diff
@@ -20,14 +20,17 @@
 use function sprintf;
 use function strtolower;
 
+/**
+ * @final
```
Ocramius, line 24: Not sure if we can add this here, unless we plan a migration to actual keyword in the next major?

src/FilterPluginManager.php
```diff
@@ -25,6 +25,9 @@
  * Enforces that filters retrieved are either callbacks or instances of
  * FilterInterface. Additionally, it registers a number of default filters
  * available, as well as aliases for them.
+ *
+ * @final
+ * @extends AbstractPluginManager<FilterInterface|callable>
```
Ocramius: `|callable` is a bit too wide, since it allows for `callable(): void`

I think `callable(): mixed` or `callable(mixed): mixed` would be more appropriate

src/FilterPluginManager.php
```diff
@@ -504,4 +509,28 @@ public function validatePlugin($plugin)
             throw new RuntimeException($e->getMessage(), $e->getCode(), $e);
         }
     }
+
+    /**
+     * @inheritDoc
+     * @param class-string<FilterInterface>|string $name Service name of plugin to retrieve.
+     * @param null|array<mixed> $options Options to use when creating the instance.
+     * @return FilterInterface|callable
```
Ocramius: All over this file, `callable` is a bit too lax

src/Inflector.php
```diff
@@ -481,7 +481,7 @@ protected function _normalizeSpec($spec)
      * Resolve named filters and convert them to filter objects.
      *
      * @param  string $rule
-     * @return FilterInterface
+     * @return FilterInterface|callable
```
Ocramius: Same note about `callable`

src/StaticFilter.php
```diff
@@ -8,7 +8,7 @@
 
 class StaticFilter
```
Ocramius, line 12: Uhh, we should really deprecate this stuff @_@

test/FilterPluginManagerTest.php
```diff
@@ -33,6 +34,7 @@ public function testFilterSuccessfullyRetrieved(): void
     public function testRegisteringInvalidFilterRaisesException(): void
     {
         $this->expectException($this->getInvalidServiceException());
+        /** @psalm-suppress InvalidArgument */
         $this->filters->setService('test', $this);
```
Ocramius, line 38: I would expect the exception right before calling `get()` - as it is written here, it is a bit unclear where it may crash

test/StaticAnalysis/PluginRetrievalTest.php
```diff
@@ -0,0 +1,34 @@
...
+
+    public function __construct(FilterPluginManager $pluginManager)
+    {
+        $this->pluginManager = $pluginManager;
+    }
+
+    /** @return mixed */
+    public function filterSomethingWithAKnownFilterClass(string $value)
+    {
+        $plugin = $this->pluginManager->get(StringToUpper::class);
```
Ocramius: Assuming type inference on the plugin manager is improved, couldn't we infer `string` as a return value here?

`StringToUpper` would return `string` if the input was a `string` :thinking:

gsteel: I started looking into adding template types to filter interface, i.e. `FilterInterface<TInput, TOutput>`, but it started getting ugly really quickly so I settled for mixed in this instance because it's intended to test the plugin manager as opposed to the filter

Ocramius: Nono, that is a mess, in this case I mean that `$plugin` could be inferred as `StringToUpper`, which has following signature:

https://github.com/laminas/laminas-filter/blob/88ed9340e0bd099ac9f5a06244ddd6e660196161/src/StringToUpper.php#L43-L46

If we document this signature as "if the input is not a scalar, return T, otherwise string" (possible with psalm conditional types), then this test would highlight how type inference goes deep here, and gives us a `string`

src/FilterChain.php
gsteel: This was an assumption that final by default would be a goal for any future major - Shall I revert?

test/StaticAnalysis/PluginRetrievalTest.php
gsteel: Got it. This would all be a whole lot easier if we only ever returned instances of FilterInterface - could we wrap all callables in the `Callback`Filter in a minor or would that require a major?

Ocramius: If `Callback` is also a `callable`, it probably wouldn't be a BC break :thinking:

test/StaticAnalysis/PluginRetrievalTest.php
```diff
@@ -0,0 +1,33 @@
...
+final class PluginRetrievalTest
+{
+    private FilterPluginManager $pluginManager;
+
+    public function __construct(FilterPluginManager $pluginManager)
+    {
+        $this->pluginManager = $pluginManager;
+    }
+
+    public function filterSomethingWithAKnownFilterClass(string $value): string
```
Ocramius, line 19: So much nicer :+1:

src/FilterChain.php
Ocramius: If we do it in the next major, we're good here :+1:

Ocramius (approved):
:ship: IMO

Good gradual improvement, ready for a release and for field testing

PR #59, opened 2022-07-12
Drop support for Service Manager 2.x series

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This patch moves the service manager dependency from `require-dev` to `require` effectively dropping support for Service Manager ^2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,12 +33,12 @@
     },
     "require": {
         "php": "^7.4 || ~8.0.0 || ~8.1.0",
+        "laminas/laminas-servicemanager": "^3.14.0",
         "laminas/laminas-stdlib": "^3.6.1"
     },
     "require-dev": {
         "laminas/laminas-coding-standard": "^2.3.0",
         "laminas/laminas-crypt": "^3.5.1",
-        "laminas/laminas-servicemanager": "^3.14.0",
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -33,12 +33,12 @@
     },
     "require": {
         "php": "^7.4 || ~8.0.0 || ~8.1.0",
+        "laminas/laminas-servicemanager": "^3.14.0",
```
Ocramius, line 36: IMO OK to introduce this, since this package ships with symbols that depend on it, but kinda sad to see that the dependency set increases :|

Ocramius (approved):

Ocramius (approved):
Let's :ship: this - we can drop this requirement later (via splitting), but it's only fair to tell our consumers that they can only use the plugin manager if they install the servicemanager.

gsteel:
@Ocramius
By splitting, do you mean splitting the package into 2, say, a collection of standalone filters and a separate package that provides plugin managers and wiring?

Ocramius:
Something like that - would hold off on that though, as it leads to maintenance effort explosion.

PR #63, opened 2022-09-09
Apply PHP 7.4 syntax and typed property

samsonasik:
Signed-off-by: Abdul Malik Ikhsan <samsonasik@gmail.com>

|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Since composer.json require php 7.4, php 7.4 syntax can be applied with typed properties. 

- For `final class`, typed properties for protected is allowed if no inherit of parent or used by its final class parent.
- Otherwise, update only private properties.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1614,13 +1614,13 @@
       <code>$matches</code>
       <code>$matches</code>
     </MissingClosureParamType>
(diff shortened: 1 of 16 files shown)


samsonasik:
All green 🎉

Ocramius (approved):
Thanks @samsonasik!

PR #64, opened 2022-09-20
The Filter Chain should be `iterable`

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes - for other components, kind of.
| BC Break      | maybe
| New Feature   | yes

### Description

Numerous components iterate over the filter chain even though it is not iterable - `laminas-form`, `laminas-inputfilter` to name 2

The class is marked soft final, so I'm unsure if this is a BC break or not.

Also bumps dev dependencies and minimum version of `StdLib` to take advantage of better types

### DIFF
--- a/src/FilterChain.php
+++ b/src/FilterChain.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Laminas\Filter;
 
 use Countable;
+use IteratorAggregate;
 use Laminas\ServiceManager\ServiceManager;
 use Laminas\Stdlib\PriorityQueue;
 use ReturnTypeWillChange;
@@ -22,8 +23,9 @@ use function strtolower;
 
 /**
  * @final
+ * @implements IteratorAggregate<array-key, FilterInterface|callable(mixed): mixed>
  */
-class FilterChain extends AbstractFilter implements Countable
+class FilterChain extends AbstractFilter implements Countable, IteratorAggregate
 {
     /**
      * Default priority at which filters are added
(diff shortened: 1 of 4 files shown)


src/FilterChain.php
```diff
@@ -225,13 +227,19 @@ public function getFilters()
      *
      * @param  mixed $value
      * @return mixed
+     * @psalm-suppress MixedAssignment values are always mixed
      */
     public function filter($value)
     {
-        $chain = clone $this->filters;
-
+        $chain         = clone $this->filters;
```
Ocramius: Couldn't this become `foreach ($this ...`? 🤔

Ocramius (changes requested):
I like the idea here - unsure if valid architecturally, since the internal details shouldn't be exposed 🤔

froschdesign:
@gsteel 
Is the same needed for the validator chain in laminas-validator?

https://github.com/laminas/laminas-validator/blob/7f15f8136f913d17ba39ee595637e1e8348b1aac/src/ValidatorChain.php#L20-L22

gsteel:
> Is the same needed for the validator chain in laminas-validator?

@froschdesign 

Yes!

Ocramius (approved):
🚢  thanks @gsteel!

gsteel:
> I like the idea here - unsure if valid architecturally, since the internal details shouldn't be exposed

Well the filter chain (and validator chain) are well known as a list of filters that are executed in order. Would it be better for `getIterator()` to return a new `ArrayIterator(iterator_to_array($this->filters))` rather than returning a clone of the `PriorityQueue`?

Ocramius:
> rather than returning a clone of the `PriorityQueue`?

That's an implementation detail: the consumer just knows it's a `Traversable`

PR #66, opened 2022-09-28
Add tests to cover `AbstractUnicode` fixing undefined offset error

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This test case has been added to document that `AbstractUnicode` lower cases it's encoding option. Also discovered that an undefined index would be caused by accessing the encoding option when no default had been set.


Ocramius (approved):
Thanks @gsteel!

PR #68, opened 2022-10-11
Fix Llaminas -> Lzf Typos

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

This appears to have gone un-noticed for a very long time. Perhaps the Lzf filter should be deprecated?


Ocramius (approved):
Thanks @gsteel

PR #73, opened 2022-11-01
Deprecate Encryption Related Filters and Adapters

gsteel:
### Description

CI is failing across the board for the OpenSSL encrypt/decrypt filter adapter.

My assumption is that the CI container was recently updated to use OpenSSL 3.x - all tests pass locally on OpenSSL 1.x

It appears that changing the algo from `RC4` to something like `AES256` and using an IV works fine, but generating and packing the IV with the result will be a BC break, as will changing the algo.

I'm out of my depth trying to get this to work without breaking BC. Ideas?

This patch is a general cleanup of the adapter and test improvements - initially I thought that the key length might be too short, hence the added keys. CI is still failing though


Ocramius:
> generating and packing the IV with the result will be a BC break, as will changing the algo.

I didn't understand why it's a BC break 🤔

src/Encrypt/Openssl.php
```diff
@@ -33,20 +33,34 @@
...
+ *     private?: string,
+ *     envelope?: string,
+ *     passphrase?: string,
+ *     package?: bool,
+ *     compression?: string|array{adapter: string},
+ * }
+ * @psalm-type CompressionOptions = array{
+ *     adapter: string,
+ * }&array<string, mixed>
+ * @final
```
Ocramius, line 53: I think we should deprecate this contraption 🤔

gsteel:
> I didn't understand why it's a BC break

Because there is no way for users to override the algo, anything encrypted with the filter and then stored for later decryption will surely fail after upgrade to a minor release - same goes for packing the IV with the encrypted payload. It's a questionable decision to use the filter like this, but otherwise, what's its use case?

src/Encrypt/Openssl.php
gsteel: Just the OpenSSL adapter or the entire encrypt/decrypt filter? Happy to bin this PR and send in a fresh one that does that, along with disabling the broken OpenSSL test case.

Ocramius: IMO the adapter at first: acting as an OpenSSL integrator is just asking for pain

Ocramius:
> there is no way for users to override the algo

I probably don't fully understand the patch, but RC4 is certainly not a viable algorithm anyway.

gsteel:
OK, I've deprecated the OpenSSL adapter and disabled tests, so this is ready for review. It's worth noting that other tests were failing too - Those that use the `BlockCipher` adapter and then OpenSSL via `laminas-crypt` - Perhaps @weierophinney can help here?

Ocramius (approved):

Ocramius:
Maybe also @ezimuel (sorry Enrico, you're still the crypto guy :P )

test/Encrypt/BlockCipherTest.php
```diff
@@ -86,11 +86,11 @@ public function testGetSetEncryption(): void
         $filter = new BlockCipherEncryption(['key' => 'testkey']);
         $filter->setVector('1234567890123456');
         $filter->setEncryption(
-            ['algorithm' => 'blowfish']
+            ['algorithm' => 'aes']
         );
         $this->assertSame(
             [
-                'algorithm'     => 'blowfish',
+                'algorithm'     => 'aes',
```
gsteel, line 93: Reason for this change: blowfish does not appear to be available on 8.1 with OpenSSL 3.x - possibly "bf" might work, but I got fed up force-pushing random strings to find something that was universally supported. I believe 'aes' is the default, hence the comment.

weierophinney (approved):
I'm good with deprecation. If you are doing encryption/decryption, doing it via a general-purpose filter is likely not a great idea, and you can easily wrap your application-specific logic in a `FilterInterface` implementation if you really need it.

Considering the fact that OpenSSL is constantly evolving, as are our various security features in PHP, I'd expect most organizations will want/need app-specific filters to ensure BC in their own applications.

gsteel:
@weierophinney - what do you think about deprecating *all* the encryption related filters and adapters, i.e. the BlockCipher adapter, the Encrypt/Decrypt Filters and also the File based encryption filters?

weierophinney:
> what do you think about deprecating _all_ the encryption related filters and adapters, i.e. the BlockCipher adapter, the Encrypt/Decrypt Filters and also the File based encryption filters?

DOOOO EEEEEEeeet!

The functionality is all present in laminas-crypt, so if folks want to do those things, they have the tools. But maintaining it here becomes difficult, and it adds a dependency to laminas-crypt (and/or specific PHP extensions) that is opt-in, and creates overhead for consumers anyways.

`FilterInterface` is dead simple, and you can use a callable if desired anyways. This gives a far easier path for application-specific cryptographic filters, and one that will be more maintainable long-term for the end-user regardless.

gsteel:
I've deprecated the remaining classes - could @weierophinney or @Ocramius give this a final once over before merge? Thanks :)

Ocramius (approved):
:ship: thanks @gsteel!

PR #77, opened 2022-11-06
General SA Improvements & Low-Hanging Fruit

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| QA            | yes

### Description

General reduction is psalm baseline - mostly corrections to documented parameter/return types, addition of array shapes for various options. Descendants of `AbstractUnicode` benefit from some extra code removal

### DIFF
--- a/src/Boolean.php
+++ b/src/Boolean.php
@@ -41,7 +41,11 @@ class Boolean extends AbstractFilter
     public const TYPE_LOCALIZED    = 256;
     public const TYPE_ALL          = 511;
 
-    /** @var array<self::TYPE_*, string> */
+    /**
+     * @deprecated since 2.26 - superseded by self::CONSTANTS
+     *
+     * @var array<self::TYPE_*, string>
+     */
     protected $constants = self::CONSTANTS;
 
     private const CONSTANTS = [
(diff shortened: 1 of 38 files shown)


src/Boolean.php
```diff
@@ -19,6 +19,14 @@
 use function sprintf;
 use function strtolower;
 
+/**
+ * @psalm-type Options = array{
+ *     type: int,
```
Ocramius: Is the type `one-of<self::TYPE_*>`?

src/ToNull.php
```diff
@@ -16,6 +16,12 @@
 use function iterator_to_array;
 use function sprintf;
 
+/**
+ * @psalm-type Options = array{
+ *     type: int,
```
Ocramius: Can we restrict this integer further?

Ocramius (changes requested):

src/Boolean.php
```diff
@@ -19,6 +19,14 @@
 use function sprintf;
 use function strtolower;
 
+/**
+ * @psalm-type Options = array{
+ *     type: int-mask-of<self::TYPE_*>,
```
Ocramius, line 24: Wait, is it a one-of, or an int-mask-of? Confusing across the file 🤔

Ocramius (changes requested):

src/Boolean.php
gsteel: Internally, it's in-mask. As a parameter, it's sometimes a string equivalent

Ocramius: Thanks for clarifying 👍

Ocramius (approved):
Most excellent, thanks @gsteel!

PR #78, opened 2022-11-07
Enable Lzf, Rar and Snappy compression tests

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Adds all the required extensions for these compression types, enables the skipped tests and fixes them along with some more SA improvements.

The extensions slow down CI a lot as expected…


.laminas-ci/pre-install.sh
```diff
@@ -0,0 +1,39 @@
...
+
+echo "extension=snappy.so" >> /etc/php/"${PHP_VERSION}"/cli/php.ini
+
+# Debug output
+php --ri lzf
+echo ""
+php --ri rar
+echo ""
+php --ri snappy
+echo ""
```
Ocramius, line 37: TBH, this seems like a rabbit hole of dependency upgrades: are we sure we want to go down the road of custom extensions? :|

Ocramius (changes requested):

.laminas-ci/pre-install.sh
gsteel: I'm totally game for deprecating all 3 for removal in 3.0 - The reason for the patch was to figure out if it was even possible for these extensions to be installed on recent versions of PHP - turns out it _is_ possible. I can't imagine use is widespread though.

Shall I deprecate these adapters now? The `rar` one is particularly bad…

Ocramius: Hmm, ok, how about this as a plan:

 1. test them now
 2. create issue for deprecation (RFC)
 3. decide what to do about them later

I'm OK with testing them as-they-are for now, but custom extensions will bite us back really hard the moment we aren't looking.

gsteel: I'll stick it on the agenda perhaps too.

Ocramius: #79 and https://github.com/laminas/technical-steering-committee/pull/128

Ocramius (approved):
:ship: this meanwhile: good improvement regardless :)

PR #94, opened 2022-12-30
add new filter to filter to an enum

reinfi:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

I added a filter to filter a given value to an enum. Is a bit like the pull request [here](https://github.com/laminas/laminas-validator/pull/168).

I'm unsure if the filter should return the original value if it is not in the enum or null. I actually added it as null because than you can be sure that your value is either null or the expected enum. 

Hopefully I added everything related to your contribution guidelines. 

Of course this filter only works > PHP 8.0. 

And I think psalm will fail for my new class, but hopefully someone of you can help me how to solve it.


src/ToEnum.php
```diff
@@ -0,0 +1,113 @@
...
+use function is_array;
+use function is_int;
+use function is_string;
+use function is_subclass_of;
+
+/**
+ * @psalm-type Options array{enum: class-string<UnitEnum>}
+ * @extends AbstractFilter<Options>
+ */
+class ToEnum extends AbstractFilter
```
Ocramius: `final` please :D

src/ToEnum.php
```diff
@@ -0,0 +1,113 @@
...
+use Traversable;
+use UnitEnum;
+
+use function is_array;
+use function is_int;
+use function is_string;
+use function is_subclass_of;
+
+/**
+ * @psalm-type Options array{enum: class-string<UnitEnum>}
```
Ocramius: I wonder if we should keep extending `AbstractFilter` at all, but that's something @gsteel should help us decide.

In this case, `new ToEnum(MyEnum::class)` seems cleaner to me, and doesn't really need the API inherited from `AbstractFilter`, although unsure if divergence from that is better or worse.

Adding `AbstractFilter` can also be done afterwards too.

src/ToEnum.php
```diff
@@ -0,0 +1,113 @@
...
+ * @extends AbstractFilter<Options>
+ */
+class ToEnum extends AbstractFilter
+{
+    /** @var Options */
+    protected $options = [
```
Ocramius: The psalm errors detected here should be similar to https://github.com/laminas/laminas-validator/pull/168

Perhaps polyfilling with a `.phpstub` file in Psalm could suffice.

src/ToEnum.php
```diff
@@ -0,0 +1,113 @@
...
+        'enum' => null,
+    ];
+
+    /**
+     * @param class-string<UnitEnum>|Traversable|Options $enumOrOptions
+     */
+    public function __construct($enumOrOptions)
```
Ocramius: I would love for this to only accept `class-string<UnitEnum>` :D

src/ToEnum.php
```diff
@@ -0,0 +1,113 @@
...
+
+        if (is_string($enumOrOptions)) {
+            $this->setEnum($enumOrOptions);
+        }
+    }
+
+    /**
+     * @param class-string<UnitEnum> $enum
+     */
+    public function setEnum(string $enum): self
```
Ocramius: IMO `setEnum()` to be dropped or made `protected`: shouldn't be called from public scope

src/ToEnum.php
Ocramius: I think this should be made more generic:

```php
/**
 * @template TEnum of UnitEnum
 */
```

Then in the constructor you can:

```php
/**
 * @param class-string<TEnum> $enum
 */
public function __construct(string $enum)
{
    // ...
}
```

This way, `filter()` becomes generic, and you can infer the correct ENUM type too:

```php
/**
 * @return TEnum|null
 */
public function filter(mixed $value): ?UnitEnum
{
    // ...
}
```

src/ToEnum.php
```diff
@@ -0,0 +1,113 @@
...
+
+        if (! is_string($value) || ! is_subclass_of($enum, 'UnitEnum')) {
+            return null;
+        }
+
+        foreach ($enum::cases() as $enumCase) {
+            if ($enumCase->name === $value) {
+                return $enumCase;
+            }
+        }
```
Ocramius: Is there a way to make this not `O(N)`?

Ocramius (commented):
Good direction! Needs some feedback from @gsteel too, but overall nice idea to start handling ENUMs directly 👍

src/ToEnum.php
reinfi: But that would not work with the factory? Or should I add a factory to handle the case? 
Extending AbstractFilter  could be easily removed because I do not really need it.

src/ToEnum.php
reinfi: I tried it but psalm will not understand my type
`psalm: UndefinedDocblockClass: Docblock-defined class, interface or enum named Laminas\Filter\TEnum does not exist`
when I use `class-string<TEnum>|Traversable|Options`. If I remove the `Traversable|Options` psalm understands it.

src/ToEnum.php
reinfi: I think I do not understand you :-(

src/ToEnum.php
```diff
@@ -0,0 +1,100 @@
...
+
+        if (is_string($enumOrOptions)) {
+            $this->setEnum($enumOrOptions);
+        }
+    }
+
+    /**
+     * @param class-string<UnitEnum> $enum
+     */
+    protected function setEnum(string $enum): self
```
gsteel: This protected method can be dropped entirely, instead calling `$this->enumClass = $whatever;` in the constructor. Also protected on final is effectively private, so if you don't want to drop it entirely, make it private.

src/ToEnum.php
```diff
@@ -0,0 +1,100 @@
...
+        return $this;
+    }
+
+    /**
+     * Defined by Laminas\Filter\FilterInterface
+     *
+     * Returns an enum representation of $value or null
+     *
+     * @param  mixed $value
+     * @return UnitEnum|null
```
gsteel: Generally speaking, most filters tend to return the input value rather than a 'failure value' such as null or false.

This doesn't hold true for _all_ filters, but I'd say definitely _most_.

I think it's worth asking @weierophinney what the general rule is here.

_(Just saw your note in the description about this. Deciding what to return is also about how the filter should or might function as part of a chain. `FilterInterface` states that an exception should be thrown if filtering is impossible. In practice, I don't think many filters behave exactly like that…)_

src/ToEnum.php
```diff
@@ -0,0 +1,100 @@
...
+            throw new RuntimeException(
+                'enum class not set'
+            );
+        }
+
+        if (! is_string($value) && ! is_int($value)) {
+            return null;
+        }
+
+        if (is_subclass_of($enum, 'BackedEnum')) {
```
gsteel:
```suggestion
        if (is_subclass_of($enum, BackedEnum::class)) {
```

Will cause psalm issues that can be baselined

src/ToEnum.php
gsteel: Perhaps

```php
if (in_array(array_column($enum::cases(), 'name'), $value, true)) {
    return constant($enum . '::' . $value);
}
```

or

```php
try {
   return constant($enum . '::' . $value);
} catch (Throwable) {
   return null;
}
```

src/ToEnum.php
```diff
@@ -0,0 +1,100 @@
...
+            return $enum::tryFrom($value);
+        }
+
+        if (! is_string($value) || ! is_subclass_of($enum, 'UnitEnum')) {
```
gsteel:
```suggestion
        if (! is_string($value) || ! is_subclass_of($enum, UnitEnum::class)) {
```

gsteel (changes requested):
👍 Nice addition :)

I'm wondering whether we can add `phpVersion=8.1` to `psalm.xml` to silence most of the psalm issues here: https://psalm.dev/docs/running_psalm/configuration/#phpversion

src/ToEnum.php
weierophinney: The filters in laminas-filter are typically _normalizers_, and the idea is that you can create a filter chain that will transform a value iteratively via multiple rules. As such, most filters will do the following:

- If they can do some normalization, they will
- Otherwise, they return the value unchanged
- If there is an error performing the operation, they raise an exception

src/ToEnum.php
reinfi: removed the method entirely, but had to inline set the type if options are `Traversable`

reinfi:
When psalm is running on php8.1 everything is now working. 
But still no idea how to handle that in the ci pipeline.

src/ToEnum.php
```diff
@@ -21,53 +25,39 @@
...
-
     /**
      * Defined by Laminas\Filter\FilterInterface
      *
-     * Returns an enum representation of $value or null
+     * Returns an enum representation of $value if matching.
      *
      * @param  mixed $value
-     * @return UnitEnum|null
      */
```
gsteel, line 60: A more precise return type is still useful here

```suggestion
     * @template TInput of mixed
     * @return UnitEnum|TInput
     */
```

Or just:

```suggestion
     * @return UnitEnum|mixed
     */
```

gsteel (changes requested):

src/ToEnum.php
gsteel: The constructor arg needs to be 

```suggestion
    public function __construct($enumOrOptions = [])
```

or

```suggestion
    public function __construct($enumOrOptions = null)
```

(null would also need to be added to the `@param` in this case)

The `FilterPluginManager` will just `new` the filter without arguments unless a factory is created _(which is pointless)_.

In order for the filter to be available in the plugin manager, `FilterPluginManager` must be updated with entries under `factories` and `aliases`

The problem is that at this point, PHPUnit will fail CI on PHP 8.0 _(Along with Psalm)_

I don't see an easy way around this at the moment unless dropping support for PHP 8.0 is an option.

Thoughts @Ocramius ?

froschdesign: > The FilterPluginManager will just new the filter without arguments unless a factory is created…

Correct! 👍🏻

src/ToEnum.php
reinfi: Hmm Codesniffer does not like it

gsteel: Probably infers mixed|mixed because on 8.0 `UnitEnum` doesn't exist (maybe). `@psalm-return` should make phpcs happy here, but psalm will still complain about `UndefinedDocblockClass`

gsteel:
Closing in favour of #240

PR #103, opened 2023-05-17
Re-organise V2 docs into a versioned subdirectory

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

Moves all docs into a versioned subdirectory in preparation of work starting on v3.


froschdesign (approved):
👍🏻

PR #111, opened 2023-11-01
Add `mixed` parameter and return types to `FilterInterface`

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

Adds parameter and return type of `mixed` to `FilterInterface::filter()` and all implementations.

Some additional property/parameter/return types have been added but avoided where possible.

The plan here is to add types everywhere in another patch later on.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,9 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="5.15.0@5c774aca4746caf3d239d9c8cadb9f882ca29352">
   <file src="src/AbstractDateDropdown.php">
-    <MissingReturnType>
-      <code>filterable</code>
-    </MissingReturnType>
     <MixedArgumentTypeCoercion>
       <code>$value</code>
     </MixedArgumentTypeCoercion>
@@ -11,15 +8,6 @@
       <code>setNullOnAllEmpty</code>
       <code>setNullOnEmpty</code>
     </PossiblyUnusedMethod>
-    <PropertyNotSetInConstructor>
-      <code>$expectedInputs</code>
-    </PropertyNotSetInConstructor>
-    <RedundantCondition>
-      <code><![CDATA[array_reduce($value, self::class . '::reduce', true)]]></code>
-    </RedundantCondition>
-    <TypeDoesNotContainType>
-      <code><![CDATA[array_reduce($value, self::class . '::reduce', false)]]></code>
-    </TypeDoesNotContainType>
   </file>
   <file src="src/AbstractFilter.php">
     <DocblockTypeContradiction>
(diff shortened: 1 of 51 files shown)


psalm-baseline.xml
```diff
@@ -718,11 +674,14 @@
     <ArgumentTypeCoercion>
       <code><![CDATA[$options['pluginManager']]]></code>
     </ArgumentTypeCoercion>
+    <InvalidArgument>
+      <code>array_values($processedParts)</code>
+    </InvalidArgument>
```
Ocramius: Anything worth investigating here?

src/AbstractDateDropdown.php
```diff
@@ -18,34 +19,29 @@
...
  */
 abstract class AbstractDateDropdown extends AbstractFilter
 {
     /**
      * If true, the filter will return null if any date field is empty
-     *
-     * @var bool
      */
-    protected $nullOnEmpty = false;
+    protected bool $nullOnEmpty = false;
```
Ocramius, line 32: Not sure if we should add native types to properties: I'm personally OK with it, but it leads to BC breaks downstream, if anybody did override this.

Ocramius: Ah, never mind, this is for the next major.

At that point, I'd just get rid of `protected`, but that's for another patch.

src/AllowList.php
```diff
@@ -85,7 +85,7 @@ public function getList()
      *
      * Will return $value if its present in the white-list. If $value is rejected then it will return null.
      */
-    public function filter($value)
+    public function filter(mixed $value): mixed
```
Ocramius, line 93: Could be refined to return one of the values of the specified list, or `null`

src/Boolean.php
```diff
@@ -206,9 +206,8 @@ public function getTranslations()
      * Returns a boolean representation of $value
      *
      * @param  null|array|bool|float|int|string $value
```
Ocramius, line 209: Lies :D

src/Decompress.php
```diff
@@ -29,10 +29,10 @@ public function __invoke($value)
...
      *
-     * @param  string $value Content to decompress
-     * @return string The decompressed content
+     * @param mixed $value Content to decompress
+     * @return mixed The decompressed content
      */
-    public function filter($value)
+    public function filter(mixed $value): mixed
     {
         if (! is_string($value) && $value !== null) {
```
Ocramius, line 39: wait, we decompress `null`? :thinking:

Ocramius (changes requested):
Some changes look like scope creep in here :D

psalm-baseline.xml
gsteel: Nah. This is psalm better understanding how little it understands this existing code. My current plan is to get native types everywhere and whittle down the baseline as much as poss for v3, so it's something I should get to later on.

src/Decompress.php
gsteel: I'll have to look closer at this later on, but I think that some of the adapters only work on files passed as options (Those adapters might be dead now), therefore there's no argument to `filter()`, or something like that…

gsteel:
@Ocramius

> Some changes look like scope creep in here :D

Not sure I understand this… This patch is just meant to nail down the contract of `filter()` to `filter(mixed): mixed` with minimal other changes. I'm planning on going through the filters after and adding BC breaking types everywhere, along with `final` and improving type inference where possible. Just trying my hardest to keep the patch size as small as possible 😬

Ocramius (approved):
@gsteel my bad: I think I initially reviewed this patch with the mindset that it was for a stable branch, but I think all is good here.

PR #113, opened 2023-11-01
Adds parameter and return types to Compress filter

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| New Feature   | yes

### Description

Adds parameter and return types to the compress filter and provides general type inference improvements and simplifications.

Needs #111

### DIFF
--- a/docs/book/v3/migration/v2-to-v3.md
+++ b/docs/book/v3/migration/v2-to-v3.md
@@ -3,8 +3,6 @@
 laminas-filter version 3 makes a number of changes that may affect your application.
 This document details those changes, and provides suggestions on how to update your application to work with version 3.
 
-## Signature Changes
-
 ## Removed Filters
 
 The following filters were deprecated in the 2.0.x series of releases and have now been removed:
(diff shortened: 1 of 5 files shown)


src/Compress/CompressionAlgorithmInterface.php
```diff
@@ -31,4 +31,7 @@ public function decompress($value);
      * @return string
      */
     public function toString();
+
+    /** @psalm-return ($option is string ? mixed : array<string, mixed>) */
+    public function getOptions(string|null $option = null): mixed;
```
Ocramius: IMO `getOptions` shouldn't be part of the interface here: getting stuff from "ready to use" services is IMO bad design.

Ocramius (changes requested):

src/Compress/CompressionAlgorithmInterface.php
gsteel: OK. I'll look at it again. Perhaps in ./src calls to ...Interface::getOptions() can be refactored, the rest of it is probably tests.

Ocramius (approved):
Thanks @gsteel!

PR #117, opened 2023-11-06
Mark all classes as `final`

gsteel:
Once this is merged, I'll work through individual filters, remove inheritance, make protected properties and methods private and add native return and parameter types…


gsteel:
Considering this approved because it is also part of #118

PR #118, opened 2023-11-06
Refactor `AllowList` Filter, Add generic template to `FilterInterface`

gsteel:
This patch is indicative of the plans for all the shipped filters where feasible:

- Removes inheritance
- Private properties instead of protected
- Native return types
- Native parameter types
- Options can no longer be Traversable

The reason options can no longer be Traversable is because we can rely on static analysis by using the documented array shape. Allowing Traversable will require a lot of type juggling and coercion, and it becomes difficult to know where to stop, for example, given the `strict` option here, previously, at least `bool|int<0,1>|'0'|'1'` were acceptable, so why not also `'y'|'n'|'yes'|'no'|'true'|'false'` too? We can all agree that native types are an improvement, so we might as well get rid of the idea that Options arrays|Traversables can have mixed values too.

#### Option value setters and getters. Can we get rid of them?

It's unlikely that filters are mutated at runtime. Most consumers will be using them via an InputFilter as part of an array spec such as:

```php
[
  'filters' => [
    [
      'name' => AllowList::class,
      'options' => [
        'strict' => true,
        'list' => ['some', 'list'],
      ],
    ],
  ],
];
```

Handling the option values in `__construct` and `setOptions` seems sufficient, and the getters just feel like a testing convenience.

We'll also still have runtime type safety via native property types.

#### Inheritance Removal…

… might be invalid in this case. I can think of 2 reasons why a parent abstract might be useful:

- To provide `__invoke(mixed):mixed` _(without useful type inference)_ that just proxies to `filter()`
- To call `setOptions` during `__construct`.

Therefore, might it be better to strip everything else out of the `AbstractFilter` bar these 2 functions?

#### Introduce an `AcceptsOptionsInterface`

something along the lines of

```php
/** @template Options of array<string, mixed> */
interface AcceptsOptionsInterface {
  
  /** @param Options $options */
  public function setOptions(array $options): self; /** preferably void */
  
}
```

This patch is cut from #117 so it'll be easier to review just https://github.com/laminas/laminas-filter/commit/fab100bcc360cc41c94e8313e66a74ad7b03e2ee in isolation.

### DIFF
--- a/src/AbstractFilter.php
+++ b/src/AbstractFilter.php
@@ -21,6 +21,7 @@ use function ucwords;
 
 /**
  * @template TOptions of array
+ * @implements FilterInterface<mixed>
  */
 abstract class AbstractFilter implements FilterInterface
 {
--- a/src/AllowList.php
+++ b/src/AllowList.php
@@ -8,7 +8,6 @@ use Laminas\Stdlib\ArrayUtils;
 
 use function array_values;
 use function in_array;
-use function is_array;
 
 /**
  * @psalm-type Options = array{
@@ -18,71 +17,16 @@ use function is_array;
  */
 final class AllowList implements FilterInterface
 {
-    private bool $strict = false;
+    private readonly bool $strict;
     /** @var list<mixed> */
-    private array $list = [];
+    private readonly array $list;
 
-    /**
-     * @param Options $options
-     */
+    /** @param Options $options */
     public function __construct(array $options = [])
     {
-        $this->setOptions($options);
-    }
-
-    /**
-     * @param Options $options
-     * @return $this
-     */
-    public function setOptions(array $options = []): self
-    {
-        $strict = $options['strict'] ?? false;
-        $list   = $options['list'] ?? [];
-
-        $this->setStrict($strict);
-        $this->setList($list);
-
-        return $this;
-    }
-
-    /**
-     * Determine whether the in_array() call should be "strict" or not. See in_array docs.
-     */
-    public function setStrict(bool $strict): void
-    {
-        $this->strict = $strict;
-    }
-
-    /**
-     * Returns whether the in_array() call should be "strict" or not. See in_array docs.
-     */
-    public function getStrict(): bool
-    {
-        return $this->strict;
-    }
-
-    /**
-     * Set the list of items to allow
-     *
-     * @param iterable<array-key, mixed> $list
-     */
-    public function setList(iterable $list = []): void
-    {
-        if (! is_array($list)) {
-            $list = ArrayUtils::iteratorToArray($list);
-        }
-
-        $this->list = array_values($list);
-    }
-
-    /**
-     * Get the list of items to allow
-     *
-     * @return list<mixed>
-     */
-    public function getList(): array
-    {
-        return $this->list;
+        $this->strict = $options['strict'] ?? false;
+        $list         = ArrayUtils::iteratorToArray($options['list'] ?? []);
+        $this->list   = array_values($list);
     }
 
     /**
--- a/src/AllowList.php
+++ b/src/AllowList.php
@@ -14,6 +14,7 @@ use function in_array;
  *     strict?: bool,
  *     list?: iterable<array-key, mixed>,
  * }
+ * @implements FilterInterface<null>
  */
 final class AllowList implements FilterInterface
 {
@@ -33,10 +34,6 @@ final class AllowList implements FilterInterface
      * {@inheritDoc}
      *
      * Will return $value if its present in the allow-list. If $value is rejected then it will return null.
-     *
-     * @template T
-     * @param T $value
-     * @return T|null
      */
     public function filter(mixed $value): mixed
     {
@@ -44,11 +41,9 @@ final class AllowList implements FilterInterface
     }
 
     /**
-     * Will return $value if its present in the allow-list. If $value is rejected then it will return null.
+     * {@inheritDoc}
      *
-     * @template T
-     * @param T $value
-     * @return T|null
+     * Will return $value if its present in the allow-list. If $value is rejected then it will return null.
      */
     public function __invoke(mixed $value): mixed
     {
--- a/src/AllowList.php
+++ b/src/AllowList.php
@@ -12,7 +12,7 @@ use function in_array;
 /**
  * @psalm-type Options = array{
  *     strict?: bool,
- *     list: iterable<array-key, mixed>,
+ *     list?: iterable<array-key, mixed>,
  * }
  * @implements FilterInterface<null>
  */
@@ -23,7 +23,7 @@ final class AllowList implements FilterInterface
     private readonly array $list;
 
     /** @param Options $options */
-    public function __construct(array $options)
+    public function __construct(array $options = [])
     {
         $this->strict = $options['strict'] ?? false;
         $list         = ArrayUtils::iteratorToArray($options['list'] ?? []);
--- a/src/AllowList.php
+++ b/src/AllowList.php
@@ -12,7 +12,7 @@ use function in_array;
 /**
  * @psalm-type Options = array{
  *     strict?: bool,
- *     list?: iterable<array-key, mixed>,
+ *     list: iterable<array-key, mixed>,
  * }
  * @implements FilterInterface<null>
  */
@@ -23,7 +23,7 @@ final class AllowList implements FilterInterface
     private readonly array $list;
 
     /** @param Options $options */
-    public function __construct(array $options = [])
+    public function __construct(array $options)
     {
         $this->strict = $options['strict'] ?? false;
         $list         = ArrayUtils::iteratorToArray($options['list'] ?? []);
--- a/src/AllowList.php
+++ b/src/AllowList.php
@@ -5,53 +5,58 @@ declare(strict_types=1);
 namespace Laminas\Filter;
 
 use Laminas\Stdlib\ArrayUtils;
-use Traversable;
 
+use function array_values;
 use function in_array;
 use function is_array;
 
 /**
  * @psalm-type Options = array{
  *     strict?: bool,
- *     list?: array,
- *     ...
+ *     list?: iterable<array-key, mixed>,
  * }
- * @extends AbstractFilter<Options>
  */
-final class AllowList extends AbstractFilter
+final class AllowList implements FilterInterface
 {
-    /** @var bool */
-    protected $strict = false;
-
-    /** @var array */
-    protected $list = [];
+    private bool $strict = false;
+    /** @var list<mixed> */
+    private array $list = [];
 
     /**
-     * @param null|array|Traversable $options
+     * @param Options $options
      */
-    public function __construct($options = null)
+    public function __construct(array $options = [])
     {
-        if (null !== $options) {
-            $this->setOptions($options);
-        }
+        $this->setOptions($options);
+    }
+
+    /**
+     * @param Options $options
+     * @return $this
+     */
+    public function setOptions(array $options = []): self
+    {
+        $strict = $options['strict'] ?? false;
+        $list   = $options['list'] ?? [];
+
+        $this->setStrict($strict);
+        $this->setList($list);
+
+        return $this;
     }
 
     /**
      * Determine whether the in_array() call should be "strict" or not. See in_array docs.
-     *
-     * @param bool $strict
      */
-    public function setStrict($strict = true): void
+    public function setStrict(bool $strict): void
     {
-        $this->strict = (bool) $strict;
+        $this->strict = $strict;
     }
 
     /**
      * Returns whether the in_array() call should be "strict" or not. See in_array docs.
-     *
-     * @return bool
      */
-    public function getStrict()
+    public function getStrict(): bool
     {
         return $this->strict;
     }
@@ -59,23 +64,23 @@ final class AllowList extends AbstractFilter
     /**
      * Set the list of items to allow
      *
-     * @param array|Traversable $list
+     * @param iterable<array-key, mixed> $list
      */
-    public function setList($list = []): void
+    public function setList(iterable $list = []): void
     {
         if (! is_array($list)) {
             $list = ArrayUtils::iteratorToArray($list);
         }
 
-        $this->list = $list;
+        $this->list = array_values($list);
     }
 
     /**
      * Get the list of items to allow
      *
-     * @return array
+     * @return list<mixed>
      */
-    public function getList()
+    public function getList(): array
     {
         return $this->list;
     }
@@ -91,6 +96,18 @@ final class AllowList extends AbstractFilter
      */
     public function filter(mixed $value): mixed
     {
-        return in_array($value, $this->getList(), $this->getStrict()) ? $value : null;
+        return in_array($value, $this->list, $this->strict) ? $value : null;
+    }
+
+    /**
+     * Will return $value if its present in the allow-list. If $value is rejected then it will return null.
+     *
+     * @template T
+     * @param T $value
+     * @return T|null
+     */
+    public function __invoke(mixed $value): mixed
+    {
+        return $this->filter($value);
     }
 }
--- a/src/FilterInterface.php
+++ b/src/FilterInterface.php
@@ -4,14 +4,26 @@ declare(strict_types=1);
 
 namespace Laminas\Filter;
 
+/** @template TFilteredValue */
 interface FilterInterface
 {
     /**
      * Returns the result of filtering $value
      *
+     * @template T
+     * @param T $value
+     * @return TFilteredValue|T
      * @throws Exception\RuntimeException If filtering $value is impossible.
      */
     public function filter(mixed $value): mixed;
 
+    /**
+     * Returns the result of filtering $value
+     *
+     * @template T
+     * @param T $value
+     * @return TFilteredValue|T
+     * @throws Exception\RuntimeException If filtering $value is impossible.
+     */
     public function __invoke(mixed $value): mixed;
 }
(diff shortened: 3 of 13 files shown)


docs/book/v3/migration/v2-to-v3.md
```diff
@@ -39,6 +39,12 @@ $filtered = $filter->filter('A String');
 
 ## Removed Features
 
+### Final by default
+
+Nearly all the shipped filters now have the final keyword applied to the class. Individual filters were not designed for inheritance, so if you have filters that do extend from any of the shipped filters, you will likely have to re-consider your design.
```
boesing: In `laminas-cache`, I marked anything final while hinting to wikipedia:
https://en.wikipedia.org/wiki/Composition_over_inheritance

Or something similar, but I tried to just let users dangle with "well, if you depend on this, then your design sucks and now you have to find a solution for yourself". 😅

src/AllowList.php
```diff
@@ -5,78 +5,82 @@
...
+        $this->setOptions($options);
+    }
 
     /**
-     * @param null|array|Traversable $options
+     * @param Options $options
+     * @return $this
      */
-    public function __construct($options = null)
+    public function setOptions(array $options = []): self
```
boesing: I'd say lets remove `setOptions` at all.
Filters should be immutable once instantiated, there is *usually* no need to modify options at runtime and thus this method should be removed.

All of the code here can be handled inside of `__construct`.

src/AllowList.php
```diff
@@ -92,6 +96,18 @@ public function getList()
...
+    }
+
+    /**
+     * Will return $value if its present in the allow-list. If $value is rejected then it will return null.
+     *
+     * @template T
+     * @param T $value
+     * @return T|null
+     */
+    public function __invoke(mixed $value): mixed
```
boesing, line 48: For what is this used for? Just curious, as I'd rather remove it.

src/AllowList.php
```diff
@@ -92,6 +96,18 @@ public function getList()
      */
     public function filter(mixed $value): mixed
     {
-        return in_array($value, $this->getList(), $this->getStrict()) ? $value : null;
+        return in_array($value, $this->list, $this->strict) ? $value : null;
+    }
+
+    /**
```
boesing, line 43: imho, public methods which are also part of the interface, should not repeat the docblock as it will lead to weird issues with static analysis in case the docblock differs.
Can we move these to the interface instead and add `@inheritdoc` (I do not add these in personal projects but I know that some codestyle is requiring this and thus I dont mind having inheritdoc around).

boesing (changes requested):
Overall, having final around works for me pretty well.

I'd remove `setOptions` and/or all the setter methods at all.

All these mutable classes are painful and when I have a look into our project, no1 ever used those setters. 

Having immutable filters makes absolutely sense to me and thus I'd be fine with changing that. But if we want to keep setter methods, lets at least remove `setOptions`, no?

docs/book/v3/migration/v2-to-v3.md
gsteel: Thanks! Just leaving a link to the [laminas-cache migration docs](https://docs.laminas.dev/laminas-cache/v3/migration/to-version-3/#breaking-changes) for myself so I can write something similar.

Ocramius (approved):

src/AllowList.php
gsteel: I'm in complete agreement here. Dropping `setOptions` would be ideal, however, I'll have to check this out - mainly, I'm not sure how the `FilterPluginManager` instantiates filters - IIRC, it's a case of

```php
$filter = new SomeFilter(); // Or build from factory
$filter->setOptions($options);
```

I could be wrong of course!

src/AllowList.php
gsteel: It's part of the `FilterInterface` contract that all filters can be invoked along with also calling `filter()` - not sure how bad of a BC break removing it would be.

src/AllowList.php
gsteel: Possibly `FilterInterface` can be improved with templates such as

https://psalm.dev/r/eb9e6c72fa

gsteel: I'll paste that inline…

```php
<?php
/**
 * @template TFiltered of mixed
 */
interface FilterInterface {
  /**
   * @return mixed|TFiltered
   */
  public function filter(mixed $input): mixed;
}

/**
 */
class StringFilter implements FilterInterface {
  /** @inheritDoc */
  public function filter(mixed $input): mixed {
    if (is_scalar($input) || $input === null) {
      return (string) $input;
    }

    return $input;
  }
}
```

src/AllowList.php
boesing: You can change the filter plugin manager as well.
AFAIK, all filters which do not have dedicated factory registered are usually depending on `InvokableFactory` which only does:

```php
new $filter($options);
```

src/AllowList.php
boesing: Hm, most probably something which is used in laminas-mvc for "convenience" in laminas-view templates...
Lets keep it as it is just a proxy anyways.
Was just curious.

src/AllowList.php
gsteel: TBH, the reason I started doing these improvements in the first place was to fit them all in to the next major whilst providing SMv4 support.

I'll amend the patch to kill off all runtime mutation of filters and then deal with the plugin manager later 👍

src/AbstractFilter.php
```diff
@@ -22,6 +22,7 @@
 
 /**
  * @template TOptions of array
+ * @implements FilterInterface<mixed>
```
gsteel, line 24: This is temporary and would normally be another template. Once I've finished with v3 changes, `AbstractFilter` will likely no longer be in use.

src/FilterInterface.php
```diff
@@ -4,14 +4,26 @@
 
 namespace Laminas\Filter;
 
+/** @template TFilteredValue */
```
gsteel, line 7: This should drastically improve inference:

```php

// /** @implements FilterInterface<string> **/
// class StringToLower implements FilterInterface {}

$filter = new StringToLower();
$filter->filter(99); // int|string
$filter->filter('Foo'); // string

```

gsteel:
Gentle nudge for review from @boesing and/or @Ocramius 🫣

src/AbstractFilter.php
Ocramius: @gsteel shouldn't this have `@implements FilterInterface<TSomethingNewForSubclassToDeclare>`, or are you saying it would make it worse?

src/AllowList.php
```diff
@@ -5,92 +5,48 @@
...
-     *
-     * @param bool $strict
-     */
-    public function setStrict($strict = true): void
+    /** @param Options $options */
+    public function __construct(array $options = [])
     {
-        $this->strict = (bool) $strict;
+        $this->strict = $options['strict'] ?? false;
+        $list         = ArrayUtils::iteratorToArray($options['list'] ?? []);
```
Ocramius, line 29: Does this filter even make sense, if the list is not provided? Would it make sense to make the option mandatory?

src/FilterInterface.php
```diff
@@ -4,14 +4,26 @@
...
 interface FilterInterface
 {
     /**
      * Returns the result of filtering $value
      *
+     * @template T
+     * @param T $value
+     * @return TFilteredValue|T
```
Ocramius, line 15: Shouldn't this be `TFilteredValue&T`?

We're trying to restrict the type, not expand it :thinking: 

This can be verified wtih `$v = (new AllowList(['list' => ['a', 'b', 'c']))->filter(123)`, which becomes `123|'a'|'b'|'c'|null` here, while we want `'a'|'b'|'c'|null` to be the actual returned type.

I believe we need to check this a bit better :mag:

Ocramius: Is every filter supposed to return `|null` BTW? Unsure about that detail...

Ocramius (changes requested):

src/AbstractFilter.php
gsteel: For the time being, adding a template there will just explode the psalm issues. Because the plan is to eradicate `AbstractFilter` in favour of `final class SomeFilter implements FilterInterface`

src/AllowList.php
gsteel: You're right - the filter doesn't make sense without a mandatory list 👍

src/FilterInterface.php
gsteel: This filter is a bit odd. Typically, filter failure returns the un-filtered value, so `$stringToLower->filter($someArray): array`. So, StringToLower for example would be `@implements FilterInterface<string>` and return `string|T` - this AllowList filter is odd because in a success scenario, it returns T vs null in a failure scenario, hence the reversal of the failure/success type.

The AllowList filter specifically could be improved with a template for the list perhaps something like:

```php
/**
 * @template MyList of list<mixed>
 * @implements FilterInterface<value-of<MyList>>
 */
final class AllowList implements FilterInterface
{
  /** @param MyList $list */
  public function __construct(private array $list) {}
}
```

And I guess we'd then need to override the filter method docs to return `TFilteredValue|null`. That would be more precise right?

I assumed that the current docs as they are here would have yielded:
```php
$v = (new AllowList(['list' => ['a', 'b', 'c']))->filter('a'); // 'a'|null
$v = (new AllowList(['list' => ['a', 'b', 'c']))->filter('z'); // 'z'|null
```

gsteel: Actually, I don't think it's possible to restrict the return type to members of the list, and it does currently restrict the type to `T|null` which accurately describes the behaviour of the filter.

Adding a template to describe the list is the only way I know of to get `@return value-of<TList>` to work, i.e. `@return value-of<$this->property>` doesn't work, _(or I'm wrong 🤷‍♂️)_ - If I use a template, you'd have to annotate every usage of the filter with `@var AllowList<list{'a'|'b'|'c'}>` right?

I might need a bit of help here to refine the filter to return anything beyond `T|null`

src/AllowList.php
gsteel: But… Options need to be optional because every filter needs to be newable without arguments

Ocramius: Sad :S

src/FilterInterface.php
Ocramius: Is `value-of<TList>` not an actual template type?

I need to think of it myself too btw, just thinking that current union type is a bit too wide

Ocramius: @gsteel does this example of yours work BTW?

```php
/**
 * @template MyList of list<mixed>
 * @implements FilterInterface<value-of<MyList>>
 */
final class AllowList implements FilterInterface
{
  /** @param MyList $list */
  public function __construct(private array $list) {}
}
```

I think this is very close to what we want, and verifiable via static analysis test code

gsteel: No - it doesn't work… I tried a bunch of things but couldn't figure out how to get inference from `value-of<$this->list>` _(value-of<$property> is not supported afaict)_, also, without annotating all of the filters you _use_ to satisfy `TList`, you end up with mixed anyhow, and a bunch of psalm issues for missing templates. It's not like we can put the template on the constructor either, because then inference is lost in the `filter` method which is where we want it - i.e. we can't do constructor property promotion because the convention of a single `$options` argument.

TBH, the other filters should be a lot more straightforward - I started with this one because it's the first one in the list alphabetically

src/AllowList.php
```diff
@@ -5,92 +5,48 @@
...
 
 /**
  * @psalm-type Options = array{
  *     strict?: bool,
- *     list?: array,
- *     ...
+ *     list?: iterable<array-key, mixed>,
  * }
- * @extends AbstractFilter<Options>
+ * @implements FilterInterface<null>
```
boesing, line 17: I am not sure why there is `null` but yet do not really have the mood to give this a more closer look.
I guess you know why and thus I trust this.

test/StaticAnalysis/AllowListChecks.php
```diff
@@ -0,0 +1,33 @@
...
+declare(strict_types=1);
+
+namespace LaminasTest\Filter\StaticAnalysis;
+
+use Laminas\Filter\AllowList;
+
+/** @psalm-suppress UnusedClass */
+final class AllowListChecks
+{
+    public function filterReturnTypeIsUnionOfInputAndNull(int $value): int|null
```
boesing, line 12: since there is `null` allowed here, I guess `null` works in the template?
so even if you remove `int` from return value here, that static analysis will still pass, right?
Not sure if that is how it supposed to be?

test/TestAsset/Alpha.php
```diff
@@ -4,21 +4,27 @@
...
+use Laminas\Filter\FilterInterface;
 
-use function is_array;
-use function is_scalar;
+use function is_string;
 use function preg_replace;
 
-/** @template-extends AbstractFilter<array{}> */
-class Alpha extends AbstractFilter
+/** @implements FilterInterface<string> */
```
boesing, line 12: I guess the type is here so that we tell what is being return from `filter` function?
If I see correct, `filter` returns `mixed` here as there is `! is_string` check which returns the value without touching it?

test/TestAsset/LowerCase.php
```diff
@@ -4,15 +4,27 @@
...
 
-use Laminas\Filter\AbstractFilter;
+use Laminas\Filter\FilterInterface;
 
+use function is_string;
 use function strtolower;
 
-/** @template-extends AbstractFilter<array{}> */
-class LowerCase extends AbstractFilter
+/** @implements FilterInterface<string> */
```
boesing, line 12: I guess the type is here so that we tell what is being return from `filter` function?
If I see correct, `filter` returns `mixed` here as there is `! is_string` check which returns the value without touching it?

test/TestAsset/StripUpperCase.php
```diff
@@ -4,15 +4,27 @@
...
 
-use Laminas\Filter\AbstractFilter;
+use Laminas\Filter\FilterInterface;
 
+use function is_string;
 use function preg_replace;
 
-/** @template-extends AbstractFilter<array{}> */
-class StripUpperCase extends AbstractFilter
+/** @implements FilterInterface<string> */
```
boesing, line 12: I guess the type is here so that we tell what is being return from `filter` function?
If I see correct, `filter` returns `mixed` here as there is `! is_string` check which returns the value without touching it?

boesing (commented):
I am not 100% sure if the templating stuff is correct here?
But I havent yet fully understood what the template type is supposed to tell us and thus, I guess thats fine? 🤷🏼‍♂️

gsteel:
Thanks for the review @boesing 

Generally speaking the template on `FilterInterface` represents the successfully filtered value.

Most filters _(But not AllowList)_, return either filtered value _(T)_ or the input _(mixed)_, so `StringToUpper` for example, would `@implements FilterInterface<uppercase-string>` and by default will return a union of `mixed|uppercase-string`.

It's only possible to narrow the return type further on `filter(mixed): mixed` with a conditional return such as `@return ($value is string ? T : mixed)`.

Because filters are not frequently used standalone, the benefit of the template isn't huge, but it does allow implementations to be precise about the return type 🤷‍♂️

I probably shouldn't have started with `AllowList` because it's different to most other filters - successful filtering returns the input and unsuccessful filtering returns `null`, so it's reversed in this case, hence `@implements TemplateInterface<null>`.

test/StaticAnalysis/AllowListChecks.php
gsteel: This test verifies that when input type is known, it will be present in the returned union. Without the `int` here, the union would be `mixed|null`

boesing: but for the allowlist there is `null` defined while it is actually the value which is returned when the filtering was not successful? even tho, the return type should be a templated type from constructor, but I guess that is not possible to declare 🤔 since return type would be value-of list option.

imho, this Filter should not exist as it combines validation with filtering to a different value. the value is converted to sth invalid when stuff is not as it should be. I would expect s1 from combining validator rather than depending on filtering. but might be a personal thing

boesing:
but for the allowlist there is null defined while it is actually the value which is returned when the filtering was not successful? even tho, the return type should be a templated type from constructor, but I guess that is not possible to declare 🤔 since return type would be value-of list option.

imho, this Filter should not exist as it combines validation with filtering to a different value. the value is converted to sth invalid when stuff is not as it should be. I would expect s1 from combining validator rather than depending on filtering. but might be a personal thing.


imho allowlist is not a good example to discuss the generics stuff as that Filter is obv special and imho should not even exist

gsteel:
I completely agree with everything you say about `AllowList` - it goes against the convention of all the other filters. What I've done here is preserve its existing behaviour, and regardless of its weirdness, what I want to do to all the other filters:

- Drop option setters and getters
- Only allow `array<string, mixed>` for defining options in the constructor
- Remove inheritance
- Where possible, make filters immutable
- native types everywhere
- Useful generics (for most filters except AllowList)

I really wish I'd started with a different filter! But 'A' came first…

boesing:
Works for me. I guess the outcome will provide some benefit, even tho the idea of generics is for type "safety" but for filtering and how the filters work, it will be always "mixed" as there is absolutely no way to tell if a filter was applied or not (unless explicit `===` comparison). 
I do not expect filters from being used directly, most of these are used in forms where neither of the generics will provide any benefit.
I still like the idea but I believe it wont provide too much benefit - but maybe u proof me wrong. 

So from me, feel free to continue the work. But keep in mind that it still has to work with laminas-form 👍🏻

PR #128, opened 2024-04-01
Factory cleanup

gsteel:
Remove references to deprecated `FactoryInterface` and affordances for ServiceManager v2.x along with general cleanup of factories and some additional type inference improvements

### DIFF
--- a/src/AbstractDateDropdown.php
+++ b/src/AbstractDateDropdown.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Laminas\Filter;
 
 use function array_reduce;
-use function assert;
 use function count;
 use function is_array;
 use function is_iterable;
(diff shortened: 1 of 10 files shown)


src/AbstractDateDropdown.php
```diff
@@ -123,6 +124,7 @@ public function filter(mixed $value): mixed
      *
      * @param array $value
      * @throws Exception\RuntimeException
+     * @psalm-assert TInput $value
```
gsteel, line 126: This is not actually sufficent for L119 above. The `@psalm-var` is still needed.

src/DateTimeSelect.php
```diff
@@ -15,12 +15,12 @@
...
- *     day: numeric,
- *     hour: numeric,
- *     minute: numeric,
- *     second: numeric
+ *     year: numeric-string,
+ *     month: numeric-string,
+ *     day: numeric-string,
+ *     hour: numeric-string,
+ *     minute: numeric-string,
+ *     second: numeric-string,
```
Ocramius: Not convinced these should be `string` :think

gsteel: 👍 - I've reverted and restricted the template also to `numeric`

Issue #110, opened 2023-10-26
[RFC]: Removal of `UriNormalizer`


boesing:
## RFC

| Q                   | A      |
| ------------------- | ------ |
| Proposed Version(s) | 3.0.0  |
| BC Break?           | Yes |

### Goal

Removal of the `UriNormalize` filter.

### Background

`laminas/laminas-uri` is marked as security only and will most probably not receive any more updates.
That package does also require `laminas/laminas-validator`, most probably to validate some stuff (haven't had a look).
I was trying to add support for `laminas/laminas-servicemanager` v4 in here and realized that there `laminas-uri` does require the service manager as well (maybe due to transitive dependency to the validator component.

To reduce the complexity of this component and to remove another component which will most probably die at some point in favor of diactoros, which does also provide URI stuff while respecting PSR-7, I think we should remove the filter from here.

### Considerations

`UriNormalization` won't be possible with v3 anymore.

### Proposal(s)

Was thinking about replacing it with diactoros but the URI component does a whole lotta normalization stuff and therefore, this that filter is totally dependent on internal `laminas-uri` logic.
IMHO, the filter should get removed and/or re-implemented in one of the upcoming v3 minors while using diactoros.
The normalization logic should be part of this library, could be migrated from `laminas-uri` for example.
I tried that for a moment but ended up requiring `laminas/laminas-escaper` as well which made me stop trying.

froschdesign:
> …realized that there `laminas-uri` does require the service manager as well (maybe due to transitive dependency to the validator component.

laminas-validator requires the service manager for the plugin manager.

laminas-uri uses validators in the following classes:

- `Laminas\Uri\Mailto`
  - `Laminas\Validator\EmailAddress`
- `Laminas\Uri\Uri`
  - `Laminas\Validator\Hostname`
  - `Laminas\Validator\Ip`
  
> …the filter should get removed and/or re-implemented in one of the upcoming v3 minors while using diactoros.

👍🏻

gsteel:
Closed via #131

PR #131, opened 2024-04-01
Remove the `UriNormalize` filter

gsteel:
Further to #110, this patch removes the filter and the dependency on `laminas-uri`.

My main argument for removal vs re-implementing is that it'll be a lot easier to re-introduce a filter that forces a string uri to use a specific scheme (And name it something like `ForceUriScheme`) than it will be to preserve BC with whatever laminas-uri was doing.

As a result we have fewer dependencies that also help pave the way to Laminas ServiceManager v4

Closes #110


froschdesign:
> And name it something like `ForceUriScheme`

We need a new issue report for this, otherwise it will be forgotten.

Ocramius (approved):

PR #132, opened 2024-04-01
Deprecate the `UriNormalize` filter

gsteel:
See #110

### DIFF
--- a/docs/book/v2/migration/preparing-for-v3.md
+++ b/docs/book/v2/migration/preparing-for-v3.md
@@ -1,0 +1,55 @@
+# Preparing for Version 3
+
+Version 3 will introduce a number of backwards incompatible changes. This document is intended to help you prepare for these changes.
+
+## Removed Features
+
+### Inheritance Changes
+
+Most filters will be closed to inheritance in v3 by employing the `final` keyword.
+To prepare for this change, search for classes in your codebase that extend from any of the concrete filters and either re-implement them completely, or consider refactoring them to use composition instead of inheritance.
+
+If you have extended an existing filter for a use-case that is not handled by this library, also consider sending a patch if you think that the library could benefit from your changes.
+
+### Compression Filter Adapter Removal
+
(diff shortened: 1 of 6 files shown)


docs/book/v2/standard-filters.md
```diff
@@ -1912,6 +1912,9 @@ as the result.
 
 ## UriNormalize
 
+CAUTION: **Deprecated**
+This filter is deprecated since version 2.36.0 and will be remove in version 3.0. There will not be a replacement.
```
froschdesign: > There will not be a replacement.

I think there should be a new filter for this?

Ocramius (approved):

docs/book/v2/standard-filters.md
gsteel: Does #133 go far enough here? If so, I'll amend to reference this one after merging

froschdesign:
@gsteel 
A deprecation requires the creation of or an entry in a migration guide. We have a good example in laminas-view: https://docs.laminas.dev/laminas-view/v2/migration/preparing-for-v3/ 😃

src/UriNormalize.php
```diff
@@ -16,6 +16,8 @@
 use function str_contains;
 
 /**
+ * @deprecated This filter will be removed in version 3.0 without replacement
```
froschdesign: This text should be the same as the deprecation message in the documentation. (see above)

docs/book/v2/migration/preparing-for-v3.md
```diff
@@ -0,0 +1,55 @@
+# Preparing for Version 3
```
froschdesign, line 1: ❤️

froschdesign (changes requested):
Great work on the migration guide! 👍🏻 

Only a minor change is required.

src/UriNormalize.php
gsteel: Thanks for the review @froschdesign

froschdesign (approved):

PR #133, opened 2024-04-01
Add a `ForceUriScheme` filter to ease removal of `UriNormalize` filter

gsteel:
Relevant to #110, #131, #132

With the planned removal of the UriNormalie filter, this new filter satisfies some of the existing functionality being removed.

### DIFF
--- a/docs/book/v2/standard-filters.md
+++ b/docs/book/v2/standard-filters.md
@@ -1181,6 +1181,25 @@ $decrypted = $filter->filter('encoded_text_normally_unreadable');
 print $decrypted;
 ```
 
+## ForceUriScheme
+
+This filter will ensure that, given a string that looks like a URI with a host-name and scheme, the scheme will be forced to `https` by default, or any other arbitrary scheme provided as an option.
+
+Any value that cannot be identified as an URI to begin with, will be returned un-filtered. Furthermore, URI parsing is rudimentary so for reliable results, you should ensure that the input is a valid URI prior to filtering.
+
+### Supported Options
+
+The only supported option is `scheme` which defaults to `https`
+
+### Example Usage
+
+```php
+$filter = new Laminas\Filter\ForceUriScheme(['scheme' => 'ftp']);
+
+$filter->filter('https://example.com/path'); // 'ftp://example.com/path'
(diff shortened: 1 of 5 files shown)


froschdesign (approved):

boesing:
hm, I do not understand why this would be needed? Do we have some kind of real-world scenario where this become useful? It feels kinda weird that a URI-like string has to be passed just to modify the scheme afterwards. 
I'd rather expect "example.org" gets "https" prepended when scheme is missing 🤔

I wonder if "parse_url" should be utilized here for the example.org but thats just an idea as I cannot see any other use-case.

gsteel:
The only point in this filter is to back fill some of the functionality that is lost by the removal of the UriNormalize filter. From what I can tell, UriNormalize used Laminas\Uri to do mostly the same thing, force a specific scheme, or add it if it was missing. Replacing the scheme when one already exists is pretty easy, but less so when there is none because at that point we need to start looking much more closely at whether the uri is valid or not.

To completely replace the functionality of the previous filter, We'd have to drag in another library such as league/uri or something.

I'm personally happy to just close this PR and have zero replacement for UriNormalize because I can't see a use-case beyond 'correcting' the input for `http://` prefixed strings, that's why I made the default replacement `https://`. I don't feel like there's much value here, but theoretically people out there were using the older `UriNormalize` and it seems reasonable to try and meet them half way.

Either way, I'll wait for @boesing approval / more discussion before merge. The important bit is that 3.0.x can now get upgraded to SMv4

froschdesign:
> It feels kinda weird that a URI-like string has to be passed just to modify the scheme afterwards.

The argument is correct because the HTML element `input` with `type` `url` uses properly-formed absolute URL: `urlscheme://restofurl`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url#value

boesing:
I was not aware that this re-introduces a feature which was available with the old filter. 
I probably would have waited for a user to request this feature just to understand the use-case. I do not have any idea where it would be beneficial to have an whatever scheme force-converted to some pre-defined scheme 😅
But since this is now existing, I do not mind having it 🤷🏼‍♂️

PR #136, opened 2024-04-03
Initial Service Manager v4 Support

gsteel:
Provides initial support for SMv4

There is still a fair bit of work to do where the FilterPluginManager is consumed by FilterChain + other filters. These will need to be refactored with the FilterPluginManager as a hard-dependency in the constructor, otherwise none of the aliases will work now they have been extracted to the config provider.

Closes #125


src/Module.php
```diff
@@ -39,4 +44,10 @@ public function init($moduleManager)
             'getFilterConfig'
         );
     }
+
+    /** @return ServiceManagerConfiguration */
+    public function getFilterConfig(): array
+    {
+        return (new ConfigProvider())()['filters'];
```
boesing: I'd rather provide a public method in `ConfigProvider` than fetching the array key but thats most probably just my personal preference.

src/ConfigProvider.php
```diff
@@ -4,42 +4,244 @@
...
      *
-     * @return array<string, mixed>
+     * @return array{dependencies: ServiceManagerConfiguration, filters: ServiceManagerConfiguration}
      */
-    public function __invoke()
+    public function __invoke(): array
     {
         return [
             'dependencies' => $this->getDependencyConfig(),
+            'filters'      => $this->filters(),
```
boesing: I explicitly kept most of the plugin manager configuration within the plugin manager when approaching migration to SM v4.
That was to prevent more BC breaks than necessary.
As you can see, you already had to write that `CreatePluginManager` helper class to instantiate an instance of the plugin manager because of this.
I do not see a real benefit by having this config in the `ConfigProvider`. 
Only reason I have this in `laminas-cache` is due to the fact that we do have satellites there, so I'd rather move this back into the plugin manager instead.

src/Module.php
froschdesign: > I'd rather provide a public method in `ConfigProvider`…

This makes sense, because it is also used elsewhere:

```php
public function getPluginManager(): FilterPluginManager
{
    $plugins = $this->plugins;
    if (! $plugins instanceof FilterPluginManager) {
        $plugins = new FilterPluginManager(new ServiceManager(), (new ConfigProvider())()['filters']);
        $this->setPluginManager($plugins);
    }

    return $plugins;
}
```

https://github.com/laminas/laminas-filter/pull/136/files#diff-3a6c9663e22963eb296b993e0b7cc8335c9f7f02f8364fc66385ed7175b5639dR124

And also in tests.

boesing (changes requested):
LGTM. Like to see first plugin manager to not implement the `AbstractSingleInstancePluginManager` class as we do have to allow `callable` and `FilterInterface` types.

However, I would keep the plugin manager configuration in the plugin manager to prevent too much of a BC break. Especially a break which is not really necessary IMO.

boesing:
I wonder if we should provide a `CallableFilter` with v4 so that we do not have this `mixed` stuff laying around.
Just not 100% sure where and how to map callables to the `CallableFilter` but that would at least reduce complexity a little bit. Just some thought which came to my mind, nothing we have to do - just wanted to drop that here.

src/Module.php
gsteel: Yes - referencing array keys is like a hidden public API and worse than simply exposing a public method 👍

src/ConfigProvider.php
gsteel: My reasoning for moving config out of the plugin manager was that all of those protected props for factories, aliases, etc are now deprecated - it made sense to me to deal with deprecations at the same time considering we'll be releasing a major.

My plan in future PRs is to introduce factories for filters that rely on the plugin manager such as `FilterChain`, where the plugin manager will be made a constructor dependency.

Do you think the best course of action is to revert or to continue with that plan?

gsteel:
> I wonder if we should provide a `CallableFilter` with v4 so that we do not have this `mixed` stuff laying around.
> Just not 100% sure where and how to map callables to the `CallableFilter` but that would at least reduce complexity a little bit. Just some thought which came to my mind, nothing we have to do - just wanted to drop that here.

I was thinking the same thing. Probably worth exploring in a separate patch

src/ConfigProvider.php
boesing: Thats why I used config constructor Argument instead. you can see that in the serializer component for example

src/Module.php
gsteel: I've removed filter config completely from the config provider in favour of a class constant on the plugin manager, as per `laminas-serializer` thanks to @boesing

src/ConfigProvider.php
gsteel: Thanks for the pointer - much better solution 👍

gsteel:
@boesing - I'd appreciate another review when you get a chance! Cheers.

boesing:
will do later, have to take a 4h train and thus need some stuff to do 😁

src/FilterChain.php
```diff
@@ -189,19 +180,13 @@ public function attach($callback, $priority = self::DEFAULT_PRIORITY)
      * Retrieves the filter from the attached plugin manager, and then calls attach()
      * with the retrieved instance.
      *
-     * @param  string $name
+     * @param class-string|string $name
```
boesing: I'd say this should be class-string which matches `InstanceType`, no?

test/FilterPluginManagerCompatibilityTest.php
```diff
@@ -5,53 +5,84 @@
...
+use Throwable;
 
+use function assert;
+use function class_exists;
+use function is_string;
 use function strpos;
 
 class FilterPluginManagerCompatibilityTest extends TestCase
 {
-    use CommonPluginManagerTrait;
```
boesing, line 20: did we drop this on purpose? is that due to the `callable` type?

gsteel: Yes, and also that trait only works with `SingleInstancePluginManager`

boesing (approved):
LGTM, just a few questions but overall pretty neat work.

gsteel:
TYVM @boesing 🙏

PR #139, opened 2024-04-04
Add deprecation notice for StaticFilter to the docs

gsteel:
This was missed in previous patches…


froschdesign:
There is also a section for the [StaticFilter on the introduction page](https://docs.laminas.dev/laminas-filter/v2/intro/#using-the-staticfilter). I would also add the message there.

froschdesign (approved):

PR #144, opened 2024-06-13
Update the Migration Guide for v3.0

gsteel:
I'll keep this PR open and update the guide as work continues on v3.0

Please feel free to chip in and make amendments/improvements as necessary!

### DIFF
--- a/docs/book/v3/migration/v2-to-v3.md
+++ b/docs/book/v3/migration/v2-to-v3.md
@@ -3,6 +3,43 @@
 laminas-filter version 3 makes a number of changes that may affect your application.
 This document details those changes, and provides suggestions on how to update your application to work with version 3.
 
+## Signature and Behaviour Changes
+
+### FilterInterface
+
+`Laminas\Filter\FilterInterface` now specifies `__invoke()` as well as `filter()` forcing all filters to also be invokable classes.
+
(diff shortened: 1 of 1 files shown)


gsteel:
@froschdesign Would you mind reviewing this in its current state so that it can be merged into 3.0.x and then updated incrementally as we work on refactoring each filter?

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -3,6 +3,47 @@
...
+    }
+    
+    public function __invoke(mixed $value) : mixed
+    {
+        return $this->filter($value); 
+    }
+}
+```
+
+Also note that the signature of both `filter()` and `__invoke()` has changed to include `mixed` as both parameter and return type, therefore, it will be necessary to add any missing parameter and return types to your custom filters.
```
froschdesign, line 36:
```suggestion
#### Change of Parameter and Return Type

Also note that the signature of both `filter()` and `__invoke()` has changed to include `mixed` as both parameter and return type, therefore, it will be necessary to add any missing parameter and return types to your custom filters.
```

froschdesign (changes requested):
I suggest to add one more headline.

froschdesign (approved):

gsteel:
Thanks @froschdesign

PR #145, opened 2024-06-13
[RFC] Drop Support for PHP 8.1 in v3.0

gsteel:
Dropping support for 8.1 will open the door to readonly classes (Yay!) and allow upgrade to PHPUnit 11 (Sort of Yay!)

Any objections?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
             "composer/package-versions-deprecated": true
         },
         "platform": {
-            "php": "8.1.99"
+            "php": "8.2.99"
         }
     },
     "extra": {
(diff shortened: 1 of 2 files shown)


weierophinney:
I'd rather see this as part of a PR that actually incorporates `readonly`, as then the justification becomes more clear. But no objections, particularly if it allows us to ship something more resilient and predictable!

gsteel:
> I'd rather see this as part of a PR that actually incorporates `readonly`, as then the justification becomes more clear. But no objections, particularly if it allows us to ship something more resilient and predictable!

I've just upgraded to SMv4 here and I'm about to refactor most of the filters to be `final readonly class SomeFilter implements FilterInterface {}` but I'll likely do it in smaller PRs to make review easier.

If you could take a look at #118 and imagine the class is also readonly there 🙏

boesing:
Having `readonly` classes vs `readonly` properties makes no difference besides syntax sugar.
If you just need `readonly` feature,  go ahead and introduce `readonly` properties instead.
I personally do not like dropping PHP versions without good reasons, so there would be a "no" from me.

gsteel:
@boesing Fair point well made… Considering the new maintenance windows for PHP it might also be a bit premature too.

PR #146, opened 2024-06-13
Remove deprecated `AbstractFilter::hasPcreUnicodeSupport` method

gsteel:
This method has been irrelevant since around PHP 5.3 ish…


boesing (approved):
LGTM. Method is deprecated anyways and thus safely to remove in next major.

PR #148, opened 2024-06-13
Refactor `BaseName` filter

gsteel:
Remove inheritance, return unfiltered values for anything except string. It does not make sense to call basename on `(string) 123` or `(string) true` for example.


Ocramius (approved):

Ocramius:
Thanks @gsteel!

PR #149, opened 2024-06-13
Refactor `Boolean` filter

gsteel:
- Removes inheritance
- Changes constructor signature to only accept a well-defined array of options
- Removes option getters and setters
- Marks class as immutable
- Fixes all psalm issues in src and tests
- Expands tests

Serves as another example for general direction of filters as per #118

In this case, assuming the template is added to `FilterInterface`, the Boolean filter would get `@implements FilterInterface<bool>`


Ocramius (approved):

Ocramius:
Thanks @gsteel!

PR #152, opened 2024-06-29
Refactor `Callback` Filter

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Removes runtime mutation of the filter
- Drops inheritance


src/Callback.php
```diff
@@ -4,110 +4,45 @@
...
-        'callback_params' => [],
-    ];
+    /** @var Closure(mixed): mixed */
+    private readonly Closure $callback;
+    private readonly array $arguments;
 
     /**
-     * @param callable|array|string|Traversable $callbackOrOptions
-     * @param array $callbackParams
+     * @param callable(mixed): mixed|Options $options
```
Ocramius: Careful with this type definition: is this interpreted as `(callable(mixed): mixed)|Options`, or `callable(mixed): (mixed|Options)`?

gsteel: 👍 Thanks - added parenthesis to clarify

Ocramius (approved):
Thanks @gsteel!

PR #154, opened 2024-06-29
Refactor Lower Case and Upper Case Filters

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Removes `AbstractUnicode` from the inheritance tree and deletes it
- Clean up `File\LowerCase`, `File\UpperCase`, `StringToLower`, `StringToUpper` and `UpperCaseWords` so that they all implement `FilterInterface` directly without any inheritance.

All the filters support an `encoding` option which is re-implemented in each class. Boilerplate is reduced by adding a new internal class `EncodingOption` for consistent handling of encoding options…

Extracted some duplicated file handling code into an internal class `FilterFileContents` that applies a filter to the contents of a file.

This removes the need for `File\LowerCase` to extend `StringToLower` for example.


src/EncodingOption.php
```diff
@@ -0,0 +1,45 @@
...
+
+use function array_map;
+use function in_array;
+use function mb_internal_encoding;
+use function mb_list_encodings;
+use function sprintf;
+use function strtolower;
+
+/** @internal */
+final class EncodingOption
```
froschdesign, line 17: This class is marked as final and internal, should the constructor then be private or is this irrelevant here?

gsteel: I don't see it as relevant - if someone wants to `new EncodingOption()` on an `@internal` class - leave them to it! I can't see a reason to prevent instantiation when it's obviously a wrapper for static methods.

froschdesign (approved):
LGTM 👍🏻

gsteel:
Thanks for the review @froschdesign :)

PR #157, opened 2024-08-09
DateTimeFormatter to check for DateTimeInterface instead of DateTime

### LINKED ISSUE #96: Check for an instance of the `DateTimeInterface` instead of `DateTime` in `DateTimeFormatter`
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.30.0

#### Summary

Why are you checking for an instance of the `DateTime` and not a `DateTimeInterface`? https://github.com/laminas/laminas-filter/blob/97e3ce0fa868567aa433ed34d6f57ee703d70d3e/src/DateTimeFormatter.php#L89
Is there some secret meaning in this?

#### Current behavior

In my application, there may be a situation when I create a `DateTimeImmutable` and pass it in a `DateTimeFormatter`, 
expected a formatted string od date/time in return, but get my object back

#### How to reproduce

```php
$filter = new DateTimeFormatter();

$date = new DateTimeImmutable($inputDate, $inputTimezone);

$start_date = $filter->filter($date->add(new \DateInterval('PT1H')));
$end_date = $filter->filter($date->add(new \DateInterval('PT7H')));
```

pinclau:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
When creating a DateTimeImutable object and passing to the DateTimeFormatter, expected format is a date/time string, but object is returned instead. 

#### How to reproduce

```php
$filter = new DateTimeFormatter();

$date = new DateTimeImmutable("2024-08-09");

$result = $filter->filter($date); 
```

### DIFF
--- a/src/DateTimeFormatter.php
+++ b/src/DateTimeFormatter.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Laminas\Filter;
 
 use DateTime;
+use DateTimeInterface;
 use Throwable;
 use Traversable;
 
@@ -87,14 +88,14 @@ class DateTimeFormatter extends AbstractFilter
             return $value;
         }
 
-        if (! is_string($value) && ! is_int($value) && ! $value instanceof DateTime) {
+        if (! is_string($value) && ! is_int($value) && ! $value instanceof DateTimeInterface) {
             return $value;
         }
 
         if (is_int($value)) {
             //timestamp
             $value = new DateTime('@' . $value);
-        } elseif (! $value instanceof DateTime) {
+        } elseif (! $value instanceof DateTimeInterface) {
             $value = new DateTime($value);
         }
 
--- a/test/DateTimeFormatterTest.php
+++ b/test/DateTimeFormatterTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace LaminasTest\Filter;
 
 use DateTime;
+use DateTimeImmutable;
 use DateTimeInterface;
 use Laminas\Filter\DateTimeFormatter;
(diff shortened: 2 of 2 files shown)


arhimede:
@gsteel  Can you take a look at this PR ?
We encounter this issue on Dotkernel side, and we decided to fix it here , in the "upstream" , instead of "downstream"

gsteel (approved):
LGTM! Thank you @cPintiuta

boesing:
Ive seen a signature in docblock which might needs to be addressed. AFAIR the filter method only accepts DateTime and could now be widened to accept the interface as well.

gsteel:
That's true, but in 3.0 the dockblock _should_ read `filter(mixed $value): T`, so whilst this is definitely valid for 2.x, is it worth addressing now? If you think so, I'll open an issue for it 👍

boesing:
Nah, just wanted to drop that as I was realizing that.
Usually, laminas introduced dedicated classes for both `DateTime` and `DateTimeImmutable`:
- https://github.com/laminas/laminas-hydrator/blob/6312d724b0aa7b0fe39b899d4d8310ea05b9f5c7/src/Strategy/DateTimeImmutableFormatterStrategy.php
- https://github.com/laminas/laminas-hydrator/blob/6312d724b0aa7b0fe39b899d4d8310ea05b9f5c7/src/Strategy/DateTimeFormatterStrategy.php


Not saying that this is the way to go but thats why I was checking the filter method on my phone which states `DateTime|int|string|mixed`.

PR #160, opened 2024-08-16
Refactor `DateTimeFormatter` Filter

### LINKED ISSUE #44: Is `DateTimeFormatter` filter missed in documentation?
Is `DateTimeFormatter` filter ( https://github.com/laminas/laminas-filter/blob/2.14.x/src/DateTimeFormatter.php ) missed in documentation? At least I wasn't able to find it here https://docs.laminas.dev/laminas-filter/  .

pinclau:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Drops inheritage
- Replaced deprecated format
- Added usage documentation


src/DateTimeFormatter.php
```diff
@@ -6,47 +6,27 @@
...
+    protected string $format = DateTimeInterface::ATOM;
 
     /**
      * Set the format string accepted by date() to use when formatting a string
-     *
-     * @param  string $format
-     * @return self
      */
-    public function setFormat($format)
+    public function setFormat(string $format): DateTimeFormatter
```
gsteel: This method should be removed completely in favour of a constructor.

The constructor should look like:

```php
public function __construct(array $options = [])
{
    $this->format = $options['format'] ?? DateTimeInterface::ATOM;
}

File: src/DateTimeFormatter.php [hunk 2]
```diff
@@ -6,47 +6,27 @@
...
-     *
-     * @param array|Traversable $options
      */
-    public function __construct($options = null)
-    {
-        if ($options !== null) {
-            $this->setOptions($options);
-        }
-    }
+    protected string $format = DateTimeInterface::ATOM;
```
gsteel:
```suggestion
    private readonly string $format;
```

src/DateTimeFormatter.php
```diff
@@ -6,47 +6,27 @@
...
 use function is_int;
 use function is_string;
 
 /**
- * @psalm-type Options = array{
- *     format?: string,
- *     ...
- * }
- * @extends AbstractFilter<Options>
+ * @implements FilterInterface<mixed>
```
gsteel:
```suggestion
 * @implements FilterInterface<string>
```

src/DateTimeFormatter.php
```diff
@@ -6,47 +6,27 @@
...
-use Traversable;
 
 use function is_int;
 use function is_string;
 
 /**
- * @psalm-type Options = array{
- *     format?: string,
- *     ...
- * }
```
gsteel: The psalm type should be retained and used as the constructor argument, but it no longer needs to be unsealed, so 

```php
/**
 *     format?: non-empty-string,
 * }
 */
```

src/DateTimeFormatter.php
```diff
@@ -78,9 +57,9 @@ public function filter(mixed $value): mixed
     /**
      * Normalize the provided value to a formatted string
      *
-     * @return string|mixed
+     * @throws Exception
      */
-    protected function normalizeDateTime(mixed $value)
+    protected function normalizeDateTime(mixed $value): mixed
```
gsteel: This method can be removed and the logic moved into `filter`. Something like:

```php
// ... Also handle int|numeric case
if ( (! is_string($value) && ! $value instanceof DateTimeInterface) || $value === '') {
    return $value;
}

try {
    $dateTime = is_string($value) ? new DateTimeImmutable($value) : $value;
} catch(Throwable $e) {
    // throw…
}

return $dateTime->format($this->format);
```

It's worth thinking about Timezone support - here. For example, should we accept a timezone option and instead of `try { new Date(whatever) }`, also attempt `DateTimeImmutable::createFromFormat($this->format, $value, $this->timezone)`

Timezone can then default to `date_default_timezone_get()`

docs/book/v3/standard-filters.md
```diff
@@ -642,6 +642,27 @@ All options can be set at instantiation or by using a related method. For exampl
 methods for `target` are `getTarget()` and `setTarget()`. You can also use the `setOptions()` method
 which accepts an array of all options.
 
+## DateTimeFormatter
+
+The filter normalize a datetime string to match the filter specified format
+
+### Supported Options
+
+There are no additional options for `Laminas\Filter\Digits`
```
gsteel: Wrong class name - also, options _are_ supported and they will need documenting

docs/book/v3/standard-filters.md
```diff
@@ -642,6 +642,27 @@ All options can be set at instantiation or by using a related method. For exampl
...
+### Basic Usage
+
+```php
+$filter = new \Laminas\Filter\DateTimeFormatter();
+
+echo $filter->filter('2024-01-01'); // => 2024-01-01T00:00:00+00:00
+echo $filter->filter(1_359_739_801); // => 2013-02-01T17:30:01+00:00
+echo $filter->filter(new DateTimeImmutable('2024-01-01')) // => 2024-01-01T00:00:00+00:00 
+
+$filter->setFormat('d-m-Y');
```
```suggestion
$filter = new \Laminas\Filter\DateTimeFormatter(['format' => 'd-m-Y']);
```

gsteel (changes requested):

src/DateTimeFormatter.php
gsteel: The template on `FilterInterface` represents the successfully filtered value. i.e. `FilterInterface::filter(mixed $value): mixed|T`

src/DateTimeFormatter.php
pinclau: Done in the latest pull request

src/DateTimeFormatter.php
pinclau: Done in the latest pull request, and i added timezone too

src/DateTimeFormatter.php
pinclau: @gsteel  Here i tried another aproach, instead of checking for different types (string, int..) and returning the values that cannot be converted to a DateTimeImmutable, i let the DateTimeImmutable try to create the date object and to catch the exception and throw an InvalidArgumentException
Also, i added the timezone support here

pinclau:
@gsteel thank you for the review, the latest pull request should reflect the changes from your review with some changes that you should review if they are ok or what need to be changed, more specific the timezone support and the `filter` method that now did not return any value that cannot be converted to a DateTimeImmutable and will throw an exception.

src/DateTimeFormatter.php
```diff
@@ -4,99 +4,68 @@
...
      */
     public function filter(mixed $value): mixed
     {
         try {
-            $result = $this->normalizeDateTime($value);
+            if (! $value instanceof DateTimeInterface) {
+                $value = is_int($value) ? '@' . (string) $value : $value;
+                $value = new DateTimeImmutable($value, $this->timezone);
+            } else {
+                $value = new DateTimeImmutable($value->format($this->format), $this->timezone);
```
gsteel: The `else` branch here is a bit confusing. I think we should drop it. Also, you cannot rely on `$value->format($this->format)` being a string that `DateTime::__construct` will understand. What if the format is `Y W`?

Assuming the filter receives a `DateTime` instance, we can leave the time-zone as-is and only use the time zone when we generate a new date from `string|int`.

We should also add tests to cover this scenario.

docs/book/v3/standard-filters.md
```diff
@@ -642,6 +642,37 @@ All options can be set at instantiation or by using a related method. For exampl
 methods for `target` are `getTarget()` and `setTarget()`. You can also use the `setOptions()` method
 which accepts an array of all options.
 
+## DateTimeFormatter
+
+The filter normalize a datetime string to match the filter specified format
```
```suggestion
This filter formats either a `DateTimeInterface` object, a string, or integer that `DateTime` will understand to a date and/or time string in the configured format.
```

docs/book/v3/standard-filters.md
```diff
@@ -642,6 +642,37 @@ All options can be set at instantiation or by using a related method. For exampl
...
+
+### Supported Options
+
+The following options are supported for `Laminas\Filter\DateTimeFormatter`
+
+- `format`: a valid date format to use when formatting a string
```
```suggestion
- `format`: a valid [date format](https://www.php.net/manual/datetime.format.php) to use when formatting a string, for example `l jS F Y`. This option defaults to `DateTimeInterface::ATOM` when unspecified
```

docs/book/v3/standard-filters.md
```diff
@@ -642,6 +642,37 @@ All options can be set at instantiation or by using a related method. For exampl
...
+- `timezone` : a valid timezone to be used when handling the date
```
```suggestion
- `timezone`: The default timezone to apply when converting a string or integer argument to a `DateTime` instance. This option falls back to the system timezone when unspecified
```

src/DateTimeFormatter.php
```diff
@@ -4,99 +4,68 @@
...
 
     /**
      * Filter a datetime string by normalizing it to the filters specified format
      *
-     * @param  DateTime|string|int|mixed $value
-     * @throws Exception\InvalidArgumentException
+     * @inheritDoc
      */
     public function filter(mixed $value): mixed
     {
```
gsteel, line 58: There should be a case here to test for completely invalid arguments and return the un-filtered values in these cases.

We can attempt to filter `non-empty-string|int|DateTimeInterface`, everything else should pass through un-filtered:

`$filter->filter([]); // []`

test/DateTimeFormatterTest.php
```diff
@@ -45,98 +27,100 @@ public static function returnUnfilteredDataProvider(): array
...
+            ['2024-31-31'],
+            [true],
+            [false],
         ];
     }
 
-    #[DataProvider('returnUnfilteredDataProvider')]
-    public function testReturnUnfiltered(mixed $input): void
+    #[DataProvider('returnExceptionDataProvider')]
+    public function testReturnExceptionForUnfilteredValues(mixed $input): void
```
gsteel: This needs reverting/splitting into 2 tests - the kinds of arguments that would remain un-filtered and the kinds of arguments that would cause an exception.

test/DateTimeFormatterTest.php
```diff
@@ -45,98 +27,100 @@ public static function returnUnfilteredDataProvider(): array
...
         $result = $filter->filter(0);
-        self::assertSame('1970-01-01T00:00:00+0000', $result);
+        self::assertSame('1970-01-01T00:00:00+00:00', $result);
     }
 
     public function testDateTimeFormatted(): void
     {
-        date_default_timezone_set('UTC');
-
         $filter = new DateTimeFormatter();
```
gsteel, line 74: The timezone should be explicitly set here so that tests don't fail on user machines in different time zones

test/DateTimeFormatterTest.php
```diff
@@ -45,98 +27,100 @@ public static function returnUnfilteredDataProvider(): array
...
+     */
     public function testSetFormat(): void
     {
-        date_default_timezone_set('UTC');
-
-        $filter = new DateTimeFormatter();
-        $filter->setFormat(DateTimeInterface::RFC1036);
-        $result = $filter->filter('2012-01-01');
-        self::assertSame('Sun, 01 Jan 12 00:00:00 +0000', $result);
+        $filter = new DateTimeFormatter([
```
gsteel, line 131: Same timezone problem here

gsteel (changes requested):

src/DateTimeFormatter.php
gsteel: I've added a review comment about this - filters are supposed to attempt to filter only arguments that it understands, so types like `null`, `array`, `float`, `stdClass` etc should be returned un-filtered rather than causing an exception.

src/DateTimeFormatter.php
```diff
@@ -4,99 +4,68 @@
...
-    }
+    private readonly DateTimeZone $timezone;
 
     /**
-     * Set the format string accepted by date() to use when formatting a string
-     *
-     * @param  string $format
-     * @return self
+     * @param Options $options
+     * @throws Exception
```
gsteel: This no longer throws

src/DateTimeFormatter.php
```diff
@@ -4,99 +4,68 @@
...
+            if (! $value instanceof DateTimeInterface) {
+                $value = is_int($value) ? '@' . (string) $value : $value;
+                $value = new DateTimeImmutable($value, $this->timezone);
+            } else {
+                $value = new DateTimeImmutable($value->format($this->format), $this->timezone);
+            }
         } catch (Throwable $e) {
-            // DateTime threw an exception, an invalid date string was provided
-            throw new Exception\InvalidArgumentException('Invalid date string provided', $e->getCode(), $e);
+            throw new InvalidArgumentException('Invalid date string provided', $e->getCode(), $e);
```
gsteel: `$e->getCode()` should be cast to `(int)` - the baseline will need refreshing too

gsteel (changes requested):

docs/book/v3/standard-filters.md
```diff
@@ -642,6 +642,37 @@ All options can be set at instantiation or by using a related method. For exampl
...
+$filter = new \Laminas\Filter\DateTimeFormatter([
+    'format' => 'd-m-Y'
+]);
+echo $filter->filter('2024-08-16 00:00:00'); // => 16-08-2024
+
+$filter = new \Laminas\Filter\DateTimeFormatter([
+    'timezone' => 'Europe/Paris'
+]);
+echo $filter->filter('2024-01-01'); // => 2024-01-01T00:00:00+01:00
+```
```
froschdesign: The code example must be split into three separate examples:

- without any options
- with format option
- with timezone option

Each example has its own heading. This improves readability.

froschdesign (changes requested):

docs/book/v3/standard-filters.md
```diff
@@ -644,29 +644,40 @@ which accepts an array of all options.
...
 The following options are supported for `Laminas\Filter\DateTimeFormatter`
 
-- `format`: a valid date format to use when formatting a string
-- `timezone` : a valid timezone to be used when handling the date
+- `format`: a valid [date format](https://www.php.net/manual/datetime.format.php) to use when formatting a string, for example `l jS F Y`. This option defaults to `DateTimeInterface::ATOM` when unspecified
+- `timezone` : The default timezone to apply when converting a string or integer argument to a `DateTime` instance. This option falls back to the system timezone when unspecified
 
 ### Basic Usage
 
+#### Without any options
```
```suggestion
#### Without Any Options
```

Titles should be title case 👍

docs/book/v3/standard-filters.md
```diff
@@ -644,29 +644,40 @@ which accepts an array of all options.
...
+
 ```php
 $filter = new \Laminas\Filter\DateTimeFormatter();
 
 echo $filter->filter('2024-01-01'); // => 2024-01-01T00:00:00+00:00
 echo $filter->filter(1_359_739_801); // => 2013-02-01T17:30:01+00:00
 echo $filter->filter(new DateTimeImmutable('2024-01-01')) // => 2024-01-01T00:00:00+00:00 
+```
 
+#### With format option
```
```suggestion
#### With `format` Option
```

docs/book/v3/standard-filters.md
```diff
@@ -644,29 +644,40 @@ which accepts an array of all options.
...
+
+```php
 $filter = new \Laminas\Filter\DateTimeFormatter([
     'format' => 'd-m-Y'
 ]);
 echo $filter->filter('2024-08-16 00:00:00'); // => 16-08-2024
+```
+
+#### With timezone option
```
```suggestion
#### With `timezone` Option
```

psalm-baseline.xml
```diff
@@ -207,12 +207,9 @@
...
   <file src="src/DateTimeFormatter.php">
-    <PossiblyInvalidArgument>
-      <code><![CDATA[$e->getCode()]]></code>
-    </PossiblyInvalidArgument>
-    <MixedArgumentTypeCoercion>
-      <code><![CDATA[$value]]></code>
-    </MixedArgumentTypeCoercion>
+    <PossiblyInvalidMethodCall>
+      <code><![CDATA[format]]></code>
+    </PossiblyInvalidMethodCall>
```
gsteel: This shouldn't be in the baseline… Where is the InvalidMethodCall?

src/DateTimeFormatter.php
```diff
@@ -35,30 +35,41 @@ final class DateTimeFormatter implements FilterInterface
...
-     * Filter a datetime string by normalizing it to the filters specified format
+     *  Filter a datetime string by normalizing it to the filters specified format
      *
      * @inheritDoc
      */
     public function filter(mixed $value): mixed
     {
+        if (
+            ! (is_string($value) && $value !== '') && ! is_int($value)
+            && ! $value instanceof DateTimeInterface
```
gsteel:
```suggestion
            ! (is_string($value) && $value !== '')
            && ! is_int($value)
            && ! $value instanceof DateTimeInterface
```

src/DateTimeFormatter.php
```diff
@@ -4,99 +4,79 @@
...
-            $value = new DateTime('@' . $value);
-        } elseif (! $value instanceof DateTimeInterface) {
-            $value = new DateTime($value);
+        try {
+            if (! $value instanceof DateTimeInterface) {
+                $value = is_int($value) ? '@' . (string) $value : $value;
+                $value = new DateTimeImmutable($value, $this->timezone);
+            }
+        } catch (Throwable $e) {
+            throw new InvalidArgumentException('Invalid date string provided');
```
gsteel:
```suggestion
        if (! $value instanceof DateTimeInterface) {
            try {
                $value = is_int($value) ? '@' . (string) $value : $value;
                $value = new DateTimeImmutable($value, $this->timezone);
            } catch (Throwable $e) {
                throw new InvalidArgumentException('Invalid date string provided');
            }
```

Nitpick - but wrapping the `try/catch` in the `if` reads more like, "If the condition is met, then try something"

test/DateTimeFormatterTest.php
```diff
@@ -73,32 +74,85 @@ public function testDateTimeFormatted(): void
 
         $filter = new DateTimeFormatter();
         $result = $filter->filter('2012-01-01');
-        self::assertSame('2012-01-01T00:00:00+0000', $result);
+        self::assertSame('2012-01-01T00:00:00+00:00', $result);
     }
 
+    /**
+     * @throws \Exception
+     */
```
gsteel: Please remove, the test doesn't throw an exception

test/DateTimeFormatterTest.php
```diff
@@ -73,32 +74,85 @@ public function testDateTimeFormatted(): void
...
+        self::expectException(Exception\InvalidArgumentException::class);
+
+        new DateTimeFormatter([
+            'timezone' => 'Continent/City',
+        ]);
+    }
+
+    /**
+     * @throws \Exception
+     */
```
gsteel: Please remove, the test doesn't throw an exception

test/DateTimeFormatterTest.php
```diff
@@ -73,32 +74,85 @@ public function testDateTimeFormatted(): void
...
 
-        $resultNewYork = $filter->filter('2012-01-01');
-        self::assertSame('2012-01-01T00:00:00-0500', $resultNewYork);
+        $resultNewYork = $filterNewYork->filter('2012-01-01');
+        self::assertSame('2012-01-01T00:00:00-05:00', $resultNewYork);
     }
 
+    /**
+     * @throws \Exception
+     */
```
gsteel, line 111: Please remove, the test doesn't throw an exception

test/DateTimeFormatterTest.php
```diff
@@ -116,27 +170,29 @@ public function testAcceptDateTimeValue(): void
 
         $filter = new DateTimeFormatter();
         $result = $filter->filter(new DateTime('2012-01-01'));
-        self::assertSame('2012-01-01T00:00:00+0000', $result);
+        self::assertSame('2012-01-01T00:00:00+00:00', $result);
     }
 
-    public function testInvalidArgumentExceptionThrownOnInvalidInput(): void
+    public function testTrowInvalidArgumentOnInvalidInput(): void
```
gsteel:
```suggestion
    public function testThrowInvalidArgumentOnInvalidInput(): void
```

gsteel (changes requested):
Just a few more tweaks!

docs/book/v3/standard-filters.md
boesing: there is actually a Website which Formats any text. 
got that from @froschdesign a few years back: https://capitalizemytitle.com/style/Chicago/

psalm-baseline.xml
pinclau: the `InvalidMethodCall` came from the last line where we format the date. `$value->format($this->format);` .Psalm saw that at that point the `$vaue` could still be an `int`. I refactored a bit that logic and i could get rid io that baseline.

src/DateTimeFormatter.php
pinclau: i refactored a little bit this logic because of the psalm

gsteel (approved):
Thanks @cPintiuta, LGTM 👍

PR #161, opened 2024-09-03
Refactor `Digits` Filter

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

- Remove inheritance and implement `FilterInterface` directly
- Remove checks for Unicode support in PCRE
- Refactor test case with additional tests

### DIFF
--- a/src/Digits.php
+++ b/src/Digits.php
@@ -4,45 +4,36 @@ declare(strict_types=1);
 
 namespace Laminas\Filter;
 
-use Laminas\Stdlib\StringUtils;
-
 use function is_float;
 use function is_int;
 use function is_string;
 use function preg_replace;
 
-/**
- * @psalm-type Options = array{}
- * @extends AbstractFilter<Options>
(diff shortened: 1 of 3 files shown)


src/Digits.php
```diff
@@ -4,45 +4,36 @@
 
 namespace Laminas\Filter;
 
-use Laminas\Stdlib\StringUtils;
```
Ocramius, line 7: Is `laminas/laminas-stdlib` still used a lot, after this change?

Ocramius (approved):

src/Digits.php
gsteel: Yes - it's still well used, but we may be able to drop it once we've got through all the filters :)

froschdesign: Correct, for example `Laminas\Stdlib\ArrayUtils::iteratorToArray`.

PR #165, opened 2024-09-04
Convert blockquotes to admonitions/call-outs

froschdesign:
Admonitions are used to highlight important information in the documentation. They are used to draw attention to specific information, such as warnings, tips, or notes. Admonitions are often used to provide additional context or to explain the implications of a particular action.

## Basic Usage

Start the admonition with one of the types in capital letters followed by a colon. The type corresponds with a specific style.
The type should be followed by the content of the admonition.

```markdown
NOTE: An example of a note.
```

### With Custom Title

```markdown
WARNING: **Plural Helper Does Not Translate**
If you need to handle both plural cases *and* translations, you must use the [`TranslatePlural` helper](https://docs.laminas.dev/laminas-i18n/view-helpers/translate-plural/); `Plural` does not translate.
```

### With Headline and Multiple Paragraphs

``````markdown
> TIP: **IDE Auto-Completion in Templates**
> The `Laminas\I18n\View\HelperTrait` trait can be used to provide auto-completion for modern IDEs. It defines the aliases of the view helpers in a DocBlock as `@method` tags.
>
> ### Usage
>
> In order to allow auto-completion in templates, `$this` variable should be type-hinted via a DocBlock at the top of a template.
> It is recommended that> always the `Laminas\View\Renderer\PhpRenderer` is added as the first type, so that the IDE can auto-suggest the default view helpers from `laminas-view`.
> The `HelperTrait` from `laminas-i18n` can be chained with a pipe symbol (a.k.a. vertical bar) `|`:
>
> ```php
> /**
>  * @var Laminas\View\Renderer\PhpRenderer|Laminas\I18n\View\HelperTrait $this
>  */
> ```
>
> The `HelperTrait` traits can be chained as many as needed, depending on which view helpers from the different Laminas component are used and where the auto-completion is to be made.
``````

### DIFF
--- a/docs/book/v3/file.md
+++ b/docs/book/v3/file.md
@@ -3,11 +3,9 @@
 laminas-filter also comes with a set of classes for filtering file contents, and
 performing file operations such as renaming.
 
-> ## $_FILES
->
-> All file filter `filter()` implementations support either a file path string
-> *or* a `$_FILES` array as the supplied argument. When a `$_FILES` array is
-> passed in, the `tmp_name` is used for the file path.
+NOTE: **`$_FILES`**
+All file filter `filter()` implementations support either a file path string *or* a `$_FILES` array as the supplied argument.
+When a `$_FILES` array is passed in, the `tmp_name` is used for the file path.
 
 ## Lowercase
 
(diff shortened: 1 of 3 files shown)


Ocramius (approved):

gsteel (approved):
LGTM 👍

PR #166, opened 2024-09-04
Add new "Basic Usage" and "Application Integration" sections

froschdesign:
Relocated "Basic Usage" and "Providing Filters via Modules" content from `intro.md` to dedicated files for better structure. Updated `mkdocs.yml` to reflect these changes, improving documentation organization and accessibility.

The application integration page may have to be completely removed later if version 3 does not work with laminas-mvc in the current version.


gsteel (approved):
LGTM!

gsteel:
> The application integration page may have to be completely removed later if version 3 does not work with laminas-mvc in the current version.

MVC support _should_ be fine as far as this component is concerned. The problem is general support for SMv4 in the various MVC components - this might get easier once the lower-level components are all upgraded, but there's still a _lot_ of work to do to make that happen…

PR #167, opened 2024-09-04
Refactor `StripNewlines` filter

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

There's not much to change for this filter because it was already pretty minimal, however, inheritance has been removed.

This patch introduces `ScalarOrArrayFilterCallback`.
 
This internal class/static method will be used to replace `AbstractFilter::applyFilterOnlyToStringableValuesAndStringableArrayValues`

Allows independent unit tests and frees filters from inheritance.

Also improves the previous implementation by recursively processing nested array structures.

### DIFF
--- a/src/ScalarOrArrayFilterCallback.php
+++ b/src/ScalarOrArrayFilterCallback.php
@@ -12,6 +12,8 @@ use function is_array;
 use function is_scalar;
 
 /**
+ * This class is internal and as such is not subject to any backwards compatibility guarantees.
+ *
  * @internal
  *
  * @psalm-internal \Laminas
--- a/src/ScalarOrArrayFilterCallback.php
+++ b/src/ScalarOrArrayFilterCallback.php
@@ -29,7 +29,7 @@ final class ScalarOrArrayFilterCallback
      * @param Closure(string): string $callback
      * @return T|string|array<array-key, string|mixed>
      */
-    public static function apply(mixed $value, Closure $callback): mixed
+    public static function applyRecursively(mixed $value, Closure $callback): mixed
     {
         if (is_scalar($value) || $value instanceof Stringable) {
             return $callback((string) $value);
@@ -37,7 +37,7 @@ final class ScalarOrArrayFilterCallback
 
         if (is_array($value)) {
             return array_map(
-                static fn (mixed $value): mixed => self::apply($value, $callback),
+                static fn (mixed $value): mixed => self::applyRecursively($value, $callback),
                 $value,
             );
         }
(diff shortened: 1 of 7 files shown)


gsteel:
Also see #168

test/StripNewlinesTest.php
```diff
@@ -4,82 +4,37 @@
...
+            'String without CRLF' => ['string', 'string'],
+            'String with CRLF'    => ["Some\nText\r\nHere", 'SomeTextHere'],
+            'Escaped CRLF'        => ['\r\n', '\r\n'],
+            'Stringable'          => [new StringableObject("Hey\nThere"), 'HeyThere'],
+            'Integer'             => [1, '1'],
+            'Float'               => [1.23, '1.23'],
+            'True'                => [true, '1'],
+            'False'               => [false, ''],
+            'Array'               => [['a' => "Hey\rThere", 'b' => null], ['a' => 'HeyThere', 'b' => null]],
+            'Nested Array'        => [['a' => ["Hey\rThere"]], ['a' => ['HeyThere']]],
```
froschdesign, line 29: Interesting! But this is not documented: https://docs.laminas.dev/laminas-filter/v2/standard-filters/#stripnewlines

froschdesign (changes requested):

src/ScalarOrArrayFilterCallback.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    /**
+     * Recursively applies a callback to an array of scalars or scalar input. Non-scalar values are skipped.
+     *
+     * @template T
+     * @param T $value
+     * @param Closure(string): string $callback
+     * @return T|string|array<array-key, string|mixed>
+     */
+    public static function apply(mixed $value, Closure $callback): mixed
```
Ocramius: I think the recursion should be made clearer in the name

src/ScalarOrArrayFilterCallback.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+final class ScalarOrArrayFilterCallback
+{
+    /**
+     * Recursively applies a callback to an array of scalars or scalar input. Non-scalar values are skipped.
+     *
+     * @template T
+     * @param T $value
+     * @param Closure(string): string $callback
+     * @return T|string|array<array-key, string|mixed>
```
Ocramius, line 30: Would add that this method is `@psalm-pure`?

Ocramius (approved):

src/ScalarOrArrayFilterCallback.php
gsteel: The given closure is possibly impure, and array_map can't be pure either I don't think.

froschdesign (approved):
LGTM 👍🏻 
Thanks!

Issue #177, opened 2024-09-08
Laminas\Filter v3 TODO


gsteel:
Work on v3 is in progress, this list is here to keep track of what's been done, and what needs to be done…

All refactored filters and components should come with entries in the migration guide and a companion patch documenting deprecations in the 2.x series where necessary.

- [x] `AbstractDateDropdown` - We should be able to remove this by refactoring inheritors #209 
- [x] `AbstractFilter` - Should be removed when possible
- [x] `AllowList` #151, #118
- [x] `BaseName` #148
- [x] `Boolean` #151, #149
- [x] `Callback` #152, #153
- [x] `Compress` and `Compress/*` #195 
- [x] `DataUnitFormatter` #155, #156
- [x] `DateSelect` #207
- [x] `DateTimeFormatter` #160
- [x] `DateTimeSelect` #209 
- [x] `Decompress` #195 
- [x] `DenyList` #174, #175, #176
- [x] `Digits` #161
- [x] `Dir` #216
- [x] `File\LowerCase` #154
- [x] `File\Rename` #222
- [x] `File\RenameUpload` #224, #223, #239 
- [x] `File\UpperCase` #154
- [x] `FilterChain` #208
- [x] `FilterPluginManager`
- [x] `FilterProviderInterface` - #204,#205
- [x] `ForceUriScheme`
- [x] `HtmlEntities` #215, #218
- [x] `Inflector` #220, #221
- [x] `MonthSelect` #200
- [x] `PregReplace` #178, #179
- [x] `RealPath` #198
- [x] `StringPrefix` #169, #170
- [x] `StringSuffix` #169, #170
- [x] `StringToLower` #154
- [x] `StringToUpper` #154
- [x] `StringTrim` #162, #163
- [x] `StripNewlines` #167
- [x] `StripTags` #214, #217
- [x] `ToFloat` #171
- [x] `ToInt` #171
- [x] `ToNull` #172, #173
- [x] `ToString` needs to apply recursively to arrays to be more useful + docs need updating #241
- [x] `UpperCaseWords` #154
- [x] `Word\AbstractSeparator` - We should be able to refactor inheritors to enable removal of this abstract class #194
- [x] `Word\CamelCaseToDash` #186
- [x] `Word\CamelCaseToSeparator` #186
- [x] `Word\CamelCaseToUnderscore` #186
- [x] `Word\DashToCamelCase` #194
- [x] `Word\DashToSeparator` #184 
[...]

PR #184, opened 2024-10-23
Refactor `DashToSeparator` Filter

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Refactoring DashToSeparator based on #177

### DIFF
--- a/docs/book/v3/word.md
+++ b/docs/book/v3/word.md
@@ -106,7 +106,7 @@ The following options are supported for `Laminas\Filter\Word\DashToSeparator`:
 ### Basic Usage
 
 ```php
-$filter = new Laminas\Filter\Word\DashToSeparator('separator' => '+');
+$filter = new Laminas\Filter\Word\DashToSeparator(['separator' => '+']);
 
 print $filter->filter('this-is-my-content');
 ```
--- a/docs/book/v3/word.md
+++ b/docs/book/v3/word.md
@@ -106,8 +106,7 @@ The following options are supported for `Laminas\Filter\Word\DashToSeparator`:
 ### Basic Usage
 
 ```php
-$filter = new Laminas\Filter\Word\DashToSeparator('+');
-// or new Laminas\Filter\Word\CamelCaseToSeparator(array('separator' => '+'));
+$filter = new Laminas\Filter\Word\DashToSeparator('separator' => '+');
 
 print $filter->filter('this-is-my-content');
 ```
--- a/src/Word/DashToSeparator.php
+++ b/src/Word/DashToSeparator.php
@@ -5,10 +5,8 @@ declare(strict_types=1);
 namespace Laminas\Filter\Word;
 
 use Laminas\Filter\FilterInterface;
+use Laminas\Filter\ScalarOrArrayFilterCallback;
 
-use function array_map;
-use function is_array;
-use function is_scalar;
 use function str_replace;
 
 /**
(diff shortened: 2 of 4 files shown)


ramchale:
Opening a Draft to seek feedback before I progress this any further. Could you let me know if this is broadly what you had in mind please?

src/Word/DashToSeparator.php
```diff
@@ -14,15 +17,42 @@
...
-            Closure::fromCallable([$this, 'filterNormalizedValue'])
+        if (! is_array($value)) {
+            if (! is_scalar($value)) {
+                return $value;
+            }
+            return $this->filterNormalizedValue((string) $value);
+        }
+
+        return $this->filterNormalizedValue(
+            array_map(static fn($item) => is_scalar($item) ? (string) $item : $item, $value)
```
ramchale: While this causes a static analysis issue, I believe it is consistent with what is being done in AbstractFilter::applyFilterOnlyToStringableValuesAndStringableArrayValues . So changing it would likely introduce a change in behaviour

src/Word/DashToSeparator.php
```diff
@@ -34,4 +64,16 @@ private function filterNormalizedValue($value)
     {
         return str_replace('-', $this->separator, $value);
     }
+
+    /** @return $this */
+    public function setSeparator(string $separator): self
```
ramchale: These accessors aren't really needed for this, but removing them would potentially impact anyone currently using this class. Any preference on keeping them or not?

gsteel: set|get Anything can be killed with 🔥

src/Word/DashToSeparator.php
gsteel: `AbstractFilter::applyFilterOnlyToStringableValuesAndStringableArrayValues` was extracted to [`ScalarOrArrayFilterCallback`](https://github.com/laminas/laminas-filter/blob/3.0.x/src/ScalarOrArrayFilterCallback.php)

In this case, I think you can just `return ScalarOrArrayFilterCallback::applyRecursively($value, fn (string $input): string => str_replace('-', $this->separator, $value));` and then drop `filterNormalizedValue`

src/Word/DashToSeparator.php
```diff
@@ -14,15 +17,42 @@
  *     ...
  * }
  * @template TOptions of Options
- * @extends AbstractSeparator<TOptions>
+ * @implements FilterInterface<mixed>
```
gsteel:
```suggestion
 * @implements FilterInterface<string|array<array-key, string|mixed>>
```

src/Word/DashToSeparator.php
```diff
@@ -14,15 +17,42 @@
...
  */
-final class DashToSeparator extends AbstractSeparator
+final class DashToSeparator implements FilterInterface
 {
+    protected string $separator = ' ';
```
gsteel:
```suggestion
    private readonly string $separator;
```

src/Word/DashToSeparator.php
```diff
@@ -14,15 +17,42 @@
...
+
+    /**
+     * @param Options|string $separator Space by default
+     */
+    public function __construct(string|array $separator = ' ')
```
gsteel: I'm going for consistent constructors throughout, so ideally:

```suggestion
    /** @param Options $separator */
    public function __construct(array $options = [])
```

The `Options` psalm type can also be sealed.

Inside the constructor, I would then apply the separator option to a private readonly property with `$this->separator = $options['separator'] ?? ' ';`

gsteel (changes requested):
Thanks @ramchale :)

Ideally, each filter refactor should be accompanied by changes to the docs too - Here's [an example](https://github.com/laminas/laminas-filter/pull/169/files) and a companion patch targeting the next 2.x minor with the relevant deprecations, like [this one](https://github.com/laminas/laminas-filter/pull/170/files)

docs/book/v3/word.md
```diff
@@ -106,8 +106,7 @@ The following options are supported for `Laminas\Filter\Word\DashToSeparator`:
 ### Basic Usage
 
 ```php
-$filter = new Laminas\Filter\Word\DashToSeparator('+');
-// or new Laminas\Filter\Word\CamelCaseToSeparator(array('separator' => '+'));
+$filter = new Laminas\Filter\Word\DashToSeparator('separator' => '+');
```
gsteel:
```suggestion
$filter = new Laminas\Filter\Word\DashToSeparator(['separator' => '+']);
```

ramchale:
Thanks. I've updated that in line with the suggestions, let me know if further changes are needed. I'll open a corresponding deprecation branch against the current minor branch when I get a little bit of time

gsteel (changes requested):
Just one tweak to the docs, otherwise LGTM!

gsteel (approved):
Thank you @ramchale :)

PR #186, opened 2024-10-23
Refactor CamelCaseToSeparator Filter

### LINKED ISSUE #9: Improve handling of digits in CamelCaseToSeparator filter
This PR aims to add the behaviour asked in #33

Since the reason behind using the CamelCaseSeparator is tokenizing strings when case is changed, usual expectation is beginning a new token when hitting a number or digit.

I will try to summarize both current and changed behaviour on numbers with some examples:

| Original Text | Current | After this PR |
| --- | --- | --- |
| Foo2016Bar |  Foo2016-Bar | Foo-2016-Bar |
| March16Events | March16-Events |  March-16-Events |
| DigitSuffix42 | Digit-Suffix42 | Digit-Suffix-42 |
| 42metersLong | 42metersLong |  42-meters-Long |

> Notice the last example: Beginning of the word with lowercase letter after `42` is also mimics a case-change. I'm still not sure is this a acceptable/desired behaviour or not.

The issue #33 is about the `CamelCaseToDashFilter` but since it derived from `CamelCaseToSeparator` like `CamelCaseToUnderscore`, this change will affect both childs of `CamelCaseToSeparator`. For this reason, I also added more tests to `CamelCaseToDashTest`.

After digging dozen of approaches including other platforms such as Java and C# I came up with this final patterns with positive lookbehind and lookaheads. (Achieving same effect with a single regex seems like impossible, at least I give up)

I'm also aware of changing an existing test is not a good practice. Since the expected behaviour is changed for digits, I forced into touching the  `CamelCaseToUnderscoreTest`:
- Pa2_Title => Pa_2_Title
- Pa2a_Title => Pa_2a_Title

Altering the `CamelCaseToUnderscoreTest` seems like mandatory if the desired behaviour will be changed.

All tests are still green.

---

Originally posted by @edigu at https://github.com/zendframework/zend-filter/pull/45

### LINKED ISSUE #10: CamelCaseToDashFilter & numbers
Not sure if this is expected behavior but seems like a bug to me, but if you pass a string with numbers into `CamelCaseToDashFilter` the dash is not applied before a number, please see below for a failing unit test.

``` php
    public function testFilterSeparatesCamelCasedWordsContainingNumbersWithDashes()
    {
        $string   = 'ItIs2016';
        $filter   = new CamelCaseToDashFilter();
        $filtered = $filter($string);
        $this->assertNotEquals($string, $filtered);
        $this->assertEquals('It-Is-2016', $filtered);
    }
```

---

Originally posted by @joshhornby at https://github.com/zendframework/zend-filter/issues/33

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Refactoring CamelCaseToSeparator based on https://github.com/laminas/laminas-filter/issues/177

### DIFF
--- a/docs/book/v3/migration/v2-to-v3.md
+++ b/docs/book/v3/migration/v2-to-v3.md
@@ -58,7 +58,10 @@ The following methods have been removed:
 - `setSeparator`
 - `getSeparator`
 
-The filter will now treat numbers as a word boundary. For example `ThisHas4Words` will filter to `This-Has-4-Words`
+The constructor now only accepts an associative array of [documented options](../word.md#camelCaseToDash).
+
+The filter will now treat numbers as a word boundary.
+For example `ThisHas4Words` will filter to `This-Has-4-Words`
 
 #### `CamelCaseToSeparator`
 
@@ -70,7 +73,10 @@ The following methods have been removed:
 - `setSeparator`
 - `getSeparator`
 
-The filter will now treat numbers as a word boundary. For example `ThisHas4Words` with the default separator will filter to `This Has 4 Words`
+The constructor now only accepts an associative array of [documented options](../word.md#camelCaseToSeparator).
+
+The filter will now treat numbers as a word boundary.
+For example `ThisHas4Words` with the default separator will filter to `This Has 4 Words`
 
 #### `CamelCaseToUnderscore`
 
@@ -82,7 +88,10 @@ The following methods have been removed:
 - `setSeparator`
 - `getSeparator`
 
-The filter will now treat numbers as a word boundary. For example `ThisHas4Words` will filter to `This_Has_4_Words`
+The constructor now only accepts an associative array of [documented options](../word.md#camelCaseToUnderscore).
+
+The filter will now treat numbers as a word boundary.
+For example `ThisHas4Words` will filter to `This_Has_4_Words`
 
 #### `DashToSeparator`
 
--- a/docs/book/v3/migration/v2-to-v3.md
+++ b/docs/book/v3/migration/v2-to-v3.md
@@ -58,6 +58,8 @@ The following methods have been removed:
 - `setSeparator`
 - `getSeparator`
 
+The filter will now treat numbers as a word boundary. For example `ThisHas4Words` will filter to `This-Has-4-Words`
+
 #### `CamelCaseToSeparator`
 
 The following methods have been removed:
@@ -68,6 +70,8 @@ The following methods have been removed:
 - `setSeparator`
 - `getSeparator`
 
+The filter will now treat numbers as a word boundary. For example `ThisHas4Words` with the default separator will filter to `This Has 4 Words`
+
 #### `CamelCaseToUnderscore`
 
 The following methods have been removed:
@@ -77,6 +81,8 @@ The following methods have been removed:
 - `isOptions`
 - `setSeparator`
 - `getSeparator`
+
+The filter will now treat numbers as a word boundary. For example `ThisHas4Words` will filter to `This_Has_4_Words`
 
 #### `DashToSeparator`
 
--- a/docs/book/v3/migration/v2-to-v3.md
+++ b/docs/book/v3/migration/v2-to-v3.md
@@ -48,6 +48,36 @@ The impact of the removal of these aliases will not affect you if you use a FQCN
 
 ### Changes to Individual Filters
 
+#### `CamelCaseToDash`
+
+The following methods have been removed:
+
+- `setOptions`
+- `getOptions`
+- `isOptions`
+- `setSeparator`
+- `getSeparator`
+
+#### `CamelCaseToSeparator`
+
+The following methods have been removed:
+
+- `setOptions`
+- `getOptions`
+- `isOptions`
+- `setSeparator`
+- `getSeparator`
+
+#### `CamelCaseToUnderscore`
+
+The following methods have been removed:
+
+- `setOptions`
+- `getOptions`
+- `isOptions`
+- `setSeparator`
+- `getSeparator`
+
 #### `DashToSeparator`
 
 The following methods have been removed:
--- a/src/Word/CamelCaseToDash.php
+++ b/src/Word/CamelCaseToDash.php
@@ -4,18 +4,26 @@ declare(strict_types=1);
 
 namespace Laminas\Filter\Word;
 
+use Laminas\Filter\FilterInterface;
+
 /**
  * @psalm-type Options = array{
  *     separator?: string,
- *     ...
  * }
  * @template TOptions of Options
- * @extends CamelCaseToSeparator<TOptions>
+ * @implements FilterInterface<string|array<array-key, string|mixed>>
  */
-final class CamelCaseToDash extends CamelCaseToSeparator
+final class CamelCaseToDash implements FilterInterface
 {
-    public function __construct()
+    public function filter(mixed $value): mixed
     {
-        parent::__construct('-');
+        $filter = new CamelCaseToSeparator(['separator' => '-']);
+
+        return $filter->filter($value);
+    }
+
+    public function __invoke(mixed $value): mixed
+    {
+        return $this->filter($value);
     }
 }
--- a/src/Word/CamelCaseToDash.php
+++ b/src/Word/CamelCaseToDash.php
@@ -6,13 +6,7 @@ namespace Laminas\Filter\Word;
 
 use Laminas\Filter\FilterInterface;
 
-/**
- * @psalm-type Options = array{
- *     separator?: string,
- * }
- * @template TOptions of Options
- * @implements FilterInterface<string|array<array-key, string|mixed>>
- */
+/** @implements FilterInterface<string|array<array-key, string|mixed>> */
 final class CamelCaseToDash implements FilterInterface
 {
     public function filter(mixed $value): mixed
--- a/src/Word/CamelCaseToSeparator.php
+++ b/src/Word/CamelCaseToSeparator.php
@@ -9,7 +9,6 @@ use Laminas\Filter\ScalarOrArrayFilterCallback;
 
 use function implode;
 use function preg_split;
-use function str_replace;
 
 use const PREG_SPLIT_DELIM_CAPTURE;
 use const PREG_SPLIT_NO_EMPTY;
@@ -54,12 +53,7 @@ final class CamelCaseToSeparator implements FilterInterface
             $value,
             fn (string $input): string => implode(
                 $this->separator,
-                preg_split(
-                    $pattern,
-                    str_replace($this->separator, '', $input),
-                    -1,
-                    PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY,
-                ),
+                preg_split($pattern, $input, -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY),
             )
         );
     }
--- a/src/Word/CamelCaseToSeparator.php
+++ b/src/Word/CamelCaseToSeparator.php
@@ -7,7 +7,12 @@ namespace Laminas\Filter\Word;
 use Laminas\Filter\FilterInterface;
 use Laminas\Filter\ScalarOrArrayFilterCallback;
 
-use function preg_replace;
+use function implode;
+use function preg_split;
+use function str_replace;
+
+use const PREG_SPLIT_DELIM_CAPTURE;
+use const PREG_SPLIT_NO_EMPTY;
 
 /**
  * @psalm-type Options = array{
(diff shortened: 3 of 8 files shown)


src/Word/CamelCaseToSeparator.php
```diff
@@ -15,23 +16,24 @@
  *     ...
```
gsteel: Please seal this type by removing the `...`

src/Word/CamelCaseToSeparator.php
```diff
@@ -15,23 +16,24 @@
...
     }
 
-    /**
-     * @param  string|string[] $value
-     * @return string|string[]
-     */
-    private function filterNormalizedValue(string|array $value): string|array
+    public function filter(mixed $value): mixed
     {
         if (StringUtils::hasPcreUnicodeSupport()) {
```
gsteel: All of this can be dropped - PHP has had PCRE unicode support since 5.1 IIRC

If you look at [issue 10](https://github.com/laminas/laminas-filter/issues/10) - there are some problems with the pattern too which I think should be fixed

```php
$pattern = '#([\p{Lu}][\p{Ll}]+)#';
$result = implode($separator, preg_split($pattern, $subject, -1, PREG_SPLIT_DELIM_CAPTURE|PREG_SPLIT_NO_EMPTY));
```

👆 That might do the trick, but it doesn't account for existing non-word chars.

src/Word/CamelCaseToSeparator.php
```diff
@@ -15,23 +16,24 @@
  *     ...
  * }
  * @template TOptions of Options
- * @extends AbstractSeparator<TOptions>
+ * @implements FilterInterface<string|array<array-key, string|mixed>>
  */
-class CamelCaseToSeparator extends AbstractSeparator
+class CamelCaseToSeparator implements FilterInterface
```
gsteel: I'd like this to be final and for the 2 descendants to compose this class instead

src/Word/CamelCaseToDash.php
```diff
@@ -16,6 +16,6 @@ final class CamelCaseToDash extends CamelCaseToSeparator
 {
     public function __construct()
```
gsteel: Perhaps drop this constructor entirely, so that we just end up with

```php
public function filter(mixed $value): mixed {
    $filter = new CamelCaseToSeparator(['separator' => '-']);
    
    return $filter->filter($value);
}
```

… and of course the required `__invoke` method and removal of  inheritance

src/Word/CamelCaseToUnderscore.php
```diff
@@ -16,6 +16,6 @@ final class CamelCaseToUnderscore extends CamelCaseToSeparator
 {
     public function __construct()
```
gsteel: Same here as with `CamelCaseToDash`

test/Word/CamelCaseToSeparatorTest.php
```diff
@@ -24,7 +24,7 @@ public function testFilterSeparatesCamelCasedWordsWithSpacesByDefault(): void
     public function testFilterSeparatesCamelCasedWordsWithProvidedSeparator(): void
```
gsteel: If possible, set up a data provider to match a decent set of cases and expected outcomes including fixes for #10 such as:

- 123LeadingNumber
- Number12InTheMiddle
- ANumberAtTheEnd42
- SomePDFFile
- Has-ExistingSeparator
- What_Happens_Here
- leadingLowerCase
- Maybe Others!

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -48,6 +48,36 @@ The impact of the removal of these aliases will not affect you if you use a FQCN
 
 ### Changes to Individual Filters
 
+#### `CamelCaseToDash`
```
gsteel, line 51: Assuming #10 is fixed, we might need an extra paragraph here to describe the behaviour change too

gsteel (changes requested):

gsteel:
@ramchale 

Had another look at the regex - I think this is more like it: https://3v4l.org/ISIYe

ramchale:
Thanks
I think the solution to #10 needs some discussion, so I've split that into it's own [Draft PR](https://github.com/laminas/laminas-filter/pull/187)

ramchale:
> @ramchale
> 
> Had another look at the regex - I think this is more like it: https://3v4l.org/ISIYe

Awesome. Hadn't seen this before I created the additional draft. I'd put in my own unicode example, but your change seems to work for that as well 😄

src/Word/CamelCaseToDash.php
```diff
@@ -4,18 +4,26 @@
...
 namespace Laminas\Filter\Word;
 
+use Laminas\Filter\FilterInterface;
+
 /**
  * @psalm-type Options = array{
  *     separator?: string,
- *     ...
  * }
  * @template TOptions of Options
```
gsteel: The type and template here are no longer necessary

src/Word/CamelCaseToUnderscore.php
```diff
@@ -4,18 +4,26 @@
...
 namespace Laminas\Filter\Word;
 
+use Laminas\Filter\FilterInterface;
+
 /**
  * @psalm-type Options = array{
  *     separator?: string,
- *     ...
  * }
  * @template TOptions of Options
```
gsteel: Same here… the type and template can go

src/Word/CamelCaseToSeparator.php
```diff
@@ -4,43 +4,42 @@
...
-
-        return preg_replace($pattern, $replacement, $value);
+        return ScalarOrArrayFilterCallback::applyRecursively(
+            $value,
+            fn (string $input): string => preg_replace(
+                ['#(?<=(?:\p{Lu}))(\p{Lu}\p{Ll})#', '#(?<=(?:\p{Ll}|\p{Nd}))(\p{Lu})#'],
+                [$this->separator . '\1', $this->separator . '\1'],
+                $input
+            )
+        );
```
gsteel:
```suggestion
    $pattern = <<<REGEXP
        /
        (
            (?:\p{Lu}\p{Ll}+) # Upper followed by lower
            |
            (?:\p{Lu}+(?!\p{Ll})) # Upper not followed by lower
            |
            (?:\p{N}+) # Runs of numbers
        )
        /ux
        REGEXP;

        return ScalarOrArrayFilterCallback::applyRecursively(
            $value,
            fn (string $input): string => implode(
                $this->separator,
                preg_split(
                    $pattern,
                    str_replace($this->separator, '', $input),
                    -1,
                    PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY,
                ),
            ),
        );
```

Something like this should work OK, obviously we need the extra test cases - not sure how to handle "invalid" string input here - i.e. where the separator or other punctuation is present. IMO, strange output for inputs such as `Foo-Bar!_Ding_Dong` is the end users fault, so perhaps drop the `str_replace` in favour of `Foo-Bar => Foo---Bar`

gsteel (changes requested):

src/Word/CamelCaseToSeparator.php
ramchale: That'd make sense. People still have the option of pre-stripping the seperator if they have a need, so might be better than forcing it

ramchale: I still need to look at the migration docs, and I'm unlikely to get back to it today, but this is where I'm at

gsteel:
@froschdesign Can I have your opinion on the use of `str_replace` here?

https://github.com/laminas/laminas-filter/pull/186#discussion_r1814793452

I can't decide if input like 'Foo-Bar' should be filtered to 'Foo-Bar' or 'Foo---Bar' - I'm currently preferring the latter because it is consistent with 'Foo/Bar' => 'Foo-/-Bar'.

WDYT?

ramchale:
> @froschdesign Can I have your opinion on the use of `str_replace` here?
> 
> [#186 (comment)](https://github.com/laminas/laminas-filter/pull/186#discussion_r1814793452)
> 
> I can't decide if input like 'Foo-Bar' should be filtered to 'Foo-Bar' or 'Foo---Bar' - I'm currently preferring the latter because it is consistent with 'Foo/Bar' => 'Foo-/-Bar'.
> 
> WDYT?

A third option is to add an entry to the constructor Options (something like `$this->deduplicateSeperator = $options['deduplicateSeperator'] ?? false;`) to toggle the use of the str_replace. Would also need to add this to CamelCaseToUnderscore and CamelCaseToDash as it currently stands.

gsteel:
> A third option is to add an entry to the constructor Options (something like `$this->deduplicateSeperator = $options['deduplicateSeperator'] ?? false;`) to toggle the use of the str_replace. Would also need to add this to CamelCaseToUnderscore and CamelCaseToDash as it currently stands.

If we're doing extra options - it would be better to perhaps to add `stripNonWord` and then preg_replace anything else with a single separator, so:

`I'mAMuppet` becomes `I-m-A-Muppet`, but at this point, we're trying to save devs from themselves - this filter is supposed to convert camelcase input, if you've not provided camelCase input then any weird output is on you, so I'm in favour of filtering `With-Dashes` to `With---Dashes` and removing any attempt to normalise input.

Also, remember that filter chain can be used, i.e. `Trim | Alnum | CamelCaseToDash`

ramchale:
Updated this in line with the conversation so far. Let me know if it needs further changes

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -48,6 +48,42 @@ The impact of the removal of these aliases will not affect you if you use a FQCN
...
+
+The following methods have been removed:
+
+- `setOptions`
+- `getOptions`
+- `isOptions`
+- `setSeparator`
+- `getSeparator`
+
+The filter will now treat numbers as a word boundary. For example `ThisHas4Words` will filter to `This-Has-4-Words`
```
froschdesign: This new behaviour is not covered in the documentation. Example:

```markdown
TIP: **New Behaviour since Version 3**
The filter will now treat numbers as a word boundary.
For example `ThisHas4Words` will filter to `This-Has-4-Words`.
```

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -48,6 +48,42 @@ The impact of the removal of these aliases will not affect you if you use a FQCN
...
+
+The following methods have been removed:
+
+- `setOptions`
+- `getOptions`
+- `isOptions`
+- `setSeparator`
+- `getSeparator`
+
+The filter will now treat numbers as a word boundary. For example `ThisHas4Words` with the default separator will filter to `This Has 4 Words`
```
froschdesign: This new behaviour is not covered in the documentation.

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -48,6 +48,42 @@ The impact of the removal of these aliases will not affect you if you use a FQCN
...
+
+The following methods have been removed:
+
+- `setOptions`
+- `getOptions`
+- `isOptions`
+- `setSeparator`
+- `getSeparator`
+
+The filter will now treat numbers as a word boundary. For example `ThisHas4Words` will filter to `This_Has_4_Words`
```
froschdesign: This new behaviour is not covered in the documentation.

docs/book/v3/migration/v2-to-v3.md
froschdesign: The note on constructor options is missing:

```markdown
The constructor now only accepts an associative array of [documented options](../word.md#camelcasetoseparator).
```

froschdesign (changes requested):
Some changes to the documentation are required.

gsteel (approved):
LGTM! Thanks @ramchale

I'll wait for @froschdesign to approve the docs changes - I think behaviour wise, what we have is fine and any weird results from users supplying input that is _not_ camel cased is the users problem.

ramchale:
@gsteel awesome, thanks. I think you solved the trickiest bit 😄

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -48,6 +48,51 @@ The impact of the removal of these aliases will not affect you if you use a FQCN
...
+
+The following methods have been removed:
+
+- `setOptions`
+- `getOptions`
+- `isOptions`
+- `setSeparator`
+- `getSeparator`
+
+The constructor now only accepts an associative array of [documented options](../word.md#dashToSeparator).
```
froschdesign:
```suggestion
The constructor now only accepts an associative array of [documented options](../word.md#camelcasetodash).
```

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -48,6 +48,51 @@ The impact of the removal of these aliases will not affect you if you use a FQCN
...
```
froschdesign:
```suggestion
The constructor now only accepts an associative array of [documented options](../word.md#camelcasetoseparator).
```

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -48,6 +48,51 @@ The impact of the removal of these aliases will not affect you if you use a FQCN
...
```
froschdesign:
```suggestion
The constructor now only accepts an associative array of [documented options](../word.md#camelcasetounderscore).
```

froschdesign (changes requested):

froschdesign:
@gsteel 
Before the release I will check and update the documentation again.

docs/book/v3/migration/v2-to-v3.md
ramchale: Thanks. Amended those

gsteel (approved):
Thanks again @ramchale 👍

PR #188, opened 2024-10-24
Add `ToString` filter

### LINKED ISSUE #98: ToString filter missing?
### Feature Request

Laminas-filter component has [ToFloat](/laminas/laminas-filter/blob/2.32.x/src/ToFloat.php), [ToInt](/laminas/laminas-filter/blob/2.32.x/src/ToInt.php) and [ToNull](/laminas/laminas-filter/blob/2.32.x/src/ToFloat.php) filters but `ToString` is missing.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

This filter should be similar as `ToInt`:

```php
class ToString extends AbstractFilter
{
    {
        if (
            is_scalar($value)
            || $value === null
            || $value instanceof Stringable
        ) {
            return (string)  $value;
        }

        return $value;
    }
}
```

Your opinions?

mlussi90:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

#98 Adding a ToString Filter according to the requirements from @vaclavvanik

### DIFF
--- a/src/FilterPluginManager.php
+++ b/src/FilterPluginManager.php
@@ -36,8 +36,6 @@ class FilterPluginManager extends AbstractPluginManager
         'Int'    => ToInt::class,
         'null'   => ToNull::class,
         'Null'   => ToNull::class,
-        'string' => ToString::class,
-        'String' => ToString::class,
 
         // I18n filters
         'alnum'        => Alnum::class,
@@ -160,9 +158,6 @@ class FilterPluginManager extends AbstractPluginManager
         'tonull'                     => ToNull::class,
         'toNull'                     => ToNull::class,
         'ToNull'                     => ToNull::class,
-        'tostring'                   => ToString::class,
-        'toString'                   => ToString::class,
-        'ToString'                   => ToString::class,
         'uppercasewords'             => UpperCaseWords::class,
         'upperCaseWords'             => UpperCaseWords::class,
         'UpperCaseWords'             => UpperCaseWords::class,
--- a/src/FilterPluginManager.php
+++ b/src/FilterPluginManager.php
@@ -32,10 +32,12 @@ class FilterPluginManager extends AbstractPluginManager
     protected $aliases = [
         // @codingStandardsIgnoreStart
         // For the future
-        'int'  => ToInt::class,
-        'Int'  => ToInt::class,
-        'null' => ToNull::class,
-        'Null' => ToNull::class,
+        'int'    => ToInt::class,
+        'Int'    => ToInt::class,
+        'null'   => ToNull::class,
+        'Null'   => ToNull::class,
+        'string' => ToString::class,
+        'String' => ToString::class,
 
         // I18n filters
         'alnum'        => Alnum::class,
@@ -158,6 +160,9 @@ class FilterPluginManager extends AbstractPluginManager
         'tonull'                     => ToNull::class,
         'toNull'                     => ToNull::class,
         'ToNull'                     => ToNull::class,
+        'tostring'                   => ToString::class,
+        'toString'                   => ToString::class,
+        'ToString'                   => ToString::class,
         'uppercasewords'             => UpperCaseWords::class,
         'upperCaseWords'             => UpperCaseWords::class,
         'UpperCaseWords'             => UpperCaseWords::class,
@@ -362,6 +367,7 @@ class FilterPluginManager extends AbstractPluginManager
         Inflector::class                   => InvokableFactory::class,
         ToInt::class                       => InvokableFactory::class,
         ToFloat::class                     => InvokableFactory::class,
+        ToString::class                    => InvokableFactory::class,
         MonthSelect::class                 => InvokableFactory::class,
         ToNull::class                      => InvokableFactory::class,
         UpperCaseWords::class              => InvokableFactory::class,
--- a/src/ToString.php
+++ b/src/ToString.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Laminas\Filter;
 
 use Stringable;
+
 use function is_scalar;
 
 final class ToString implements FilterInterface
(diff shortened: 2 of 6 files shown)


src/ToString.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Filter;
+
+use function is_scalar;
+
+class ToString extends AbstractFilter
```
gsteel:
```suggestion
final class ToString implements FilterInterface
```

src/ToString.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    /**
+     * Returns (string) $value
+     *
+     * If the value provided is non-scalar, the value will remain unfiltered
+     *
+     * @param mixed $value
+     * @psalm-return ($value is scalar ? string : mixed)
+     */
+    public function filter($value): mixed
```
gsteel:
```suggestion
     * @return ($value is scalar ? string : mixed)
     */
    public function filter(mixed $value): mixed
```

src/ToString.php
```diff
@@ -0,0 +1,27 @@
...
+    {
+        if (! is_scalar($value)) {
+            return $value;
+        }
+
+        return (string) $value;
+    }
```
gsteel, line 30:
```suggestion
    }
    
    public function __invoke(mixed $value): mixed
    {
        return $this->filter($value);
    }
```

test/ToStringTest.php
```diff
@@ -0,0 +1,55 @@
...
+namespace LaminasTest\Filter;
+
+use Laminas\Filter\ToString;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+use stdClass;
+
+class ToStringTest extends TestCase
+{
+    /** @return list<array{0: mixed}> */
```
gsteel:
```suggestion
    /** @return list<array{0: mixed, 1: string}> */
```

src/ToString.php
```diff
@@ -0,0 +1,27 @@
...
+     * Returns (string) $value
+     *
+     * If the value provided is non-scalar, the value will remain unfiltered
+     *
+     * @param mixed $value
+     * @psalm-return ($value is scalar ? string : mixed)
+     */
+    public function filter($value): mixed
+    {
+        if (! is_scalar($value)) {
```
gsteel: This condition should include `Stringable`

test/ToStringTest.php
```diff
@@ -0,0 +1,55 @@
...
+    public static function returnBasicDataProvider(): array
```
gsteel, line 16: Data providers always appear as unused by Psalm when only referenced by the `DataProvider` attribute.

Once all other psalm issues are fixed, these 'PossiblyUnusedMethod' issues can be added to the baseline with `psalm --set-baseline=psalm-baseline.xml`

test/ToStringTest.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
gsteel, line 1: Will also need a test to cover added `__invoke` method and data provider cases covering `Stringable` objects

gsteel (changes requested):
Thanks @mlussi90 

The `ToString` filter will need to be added to the default set of filters in `FilterPluginManager` - additionally, the v2 docs will need updating 👍

mlussi90:
i tried to implement all your requested changes, can you take a look?

src/FilterPluginManager.php
```diff
@@ -32,10 +32,12 @@ class FilterPluginManager extends AbstractPluginManager
...
-        'int'  => ToInt::class,
-        'Int'  => ToInt::class,
-        'null' => ToNull::class,
-        'Null' => ToNull::class,
+        'int'    => ToInt::class,
+        'Int'    => ToInt::class,
+        'null'   => ToNull::class,
+        'Null'   => ToNull::class,
+        'string' => ToString::class,
+        'String' => ToString::class,
```
gsteel: I'd rather not see any new aliases please - It's better that users get into the habit of referring to filters by the FQCN, i.e. `\Laminas\Filter\ToString::class`

src/FilterPluginManager.php
```diff
@@ -158,6 +160,9 @@ class FilterPluginManager extends AbstractPluginManager
         'tonull'                     => ToNull::class,
         'toNull'                     => ToNull::class,
         'ToNull'                     => ToNull::class,
+        'tostring'                   => ToString::class,
+        'toString'                   => ToString::class,
+        'ToString'                   => ToString::class,
```
gsteel: Same comment about aliases here too. Please remove.

src/FilterPluginManager.php
```diff
@@ -362,6 +367,7 @@ class FilterPluginManager extends AbstractPluginManager
         Inflector::class                   => InvokableFactory::class,
         ToInt::class                       => InvokableFactory::class,
         ToFloat::class                     => InvokableFactory::class,
+        ToString::class                    => InvokableFactory::class,
```
gsteel, line 365: This should be all we need 👍

test/ToStringTest.php
```diff
@@ -0,0 +1,65 @@
...
+
+        self::assertSame($output, $filter->filter($input));
+    }
+
+    #[DataProvider('returnBasicDataProvider')]
+    public function testInvoke(mixed $input, string $output): void
+    {
+        $filter = new ToString();
+
+        self::assertSame($output, $filter($input));
```
gsteel:
```suggestion
        self::assertSame($output, $filter->__invoke($input));
```

Sorry - a bit nit picky, but explicitly calling `__invoke` will ensure that the call cannot be interpreted any other way.

gsteel (changes requested):
Thanks @mlussi90 - Nearly there…

You can fix coding style problems by running `vendor/bin/phpcbf` and psalm issues with the tests can be silenced with `vendor/bin/psalm --set-baseline=psalm-baseline.xml`

gsteel (changes requested):
Thank you @mlussi90 

In order to get CI green for Psalm, you'll need to refresh the baseline - this sort of thing can happen, particularly on Mac, so it's not your fault!

Try running `vendor/bin/psalm --update-baseline --no-cache --threads=1` and verify there are no more problems with `vendor/bin/psalm --no-cache --threads=1`, then commit the altered baseline.

The BC check failure can be ignored - it's complaining about the altered default value for plugin manager factories.

mlussi90:
@gsteel thank you very much for your help, the checks are looking better now i think

gsteel (approved):
Thanks @mlussi90 

I'll merge and release shortly, once I've got the docs organised 👍

PR #195, opened 2024-11-08
Refactor/Replace Compression & Decompression Filters

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | not yet
| BC Break      | yes

### Description

The old `Compress` and `Decompress` filters were a mess. Different adapters had different or unknown capabilities. Disregarding the now removed formats such as rar and lzf, the Tar and Zip adapters cannot simply compress arbitrary strings but can of course be used to compress and decompress files/archives, and the BZ2 and GZ adapters aren't suited to working with archives but are capable of compressing arbitary strings.

Instead of attempting to do all of these things with all the adapters, I've split the adapters into 2 interfaces:

- `ArchiveAdapterInterface`
- `StringCompressionAdapterInterface`

and replaced the `Compress` and `Decompress` filters with

- `CompressString`
- `CompressToArchive`
- `DecompressString`
- `DecompressArchive`

The file related filter will now accept paths, directories, PSR uploads and PHP `$_FILES` arrays.

The tools added here will also be useful for the other file related filters.

I've introduced the concept of an `ArchiveAdapterResolver`, shipping concrete implementations by default, that first scan the mime-type of the file to see if we have a matching adapter and then guess the filetype from the filename extension.

TODO:

- [ ] Documentation and migration
- [x] A PR for the deprecations


gsteel:
@froschdesign - I'd appreciate a review on the general direction of this patch when you get the chance, before I do more work on the outstanding items.

Reviews welcome from anyone else too :)

cc @alexmerlin, @ramchale, @Xerkus, @samsonasik

src/Compress/Bz2.php
```diff
@@ -4,183 +4,66 @@
...
  *
  * @psalm-type Options = array{
- *     blocksize?: int,
- *     archive?: string|null,
+ *     blocksize?: int<1, 9>|null,
  * }
- * @extends AbstractCompressionAlgorithm<Options>
  */
-final class Bz2 extends AbstractCompressionAlgorithm
+final class Bz2 implements StringCompressionAdapterInterface
```
froschdesign: @gsteel 
They are not filters, but adapters for filters. Unfortunately, this is not clear from the name of the class or namespace. Only the interface indicates this.
I think it would be helpful to make this clearer somehow.

gsteel: Considering they are complete rewrites too, I agree that renaming them is probably a good idea 👍

test/File/FileInformationTest.php
```diff
@@ -0,0 +1,107 @@
...
+
+        $upload = new UploadedFile(
+            $path,
+            filesize($path),
+            UPLOAD_ERR_OK,
+            'Foo.txt',
+            'text/plain',
+        );
+
+        $file = FileInformation::factory($upload);
```
ramchale, line 66: If you want to avoid the Laminas\Diactoros\UploadedFile import
```suggestion
                $mockStream = $this->createMock(StreamInterface::class);
        $mockStream->expects(self::once())->method('getMetadata')->with('uri')->willReturn($path);

        $mockFileUpload = $this->createMock(UploadedFileInterface::class);
        $mockFileUpload->expects(self::once())->method('getStream')->willReturn($mockStream);
        $mockFileUpload->expects(self::once())->method('getClientFilename')->willReturn('Foo.txt');
        $mockFileUpload->expects(self::once())->method('getClientMediaType')->willReturn('text/plain');
        
        $file = FileInformation::factory($mockFileUpload);
```

test/CompressToArchiveTest.php
```diff
@@ -0,0 +1,196 @@
...
+        $upload = new UploadedFile(
+            $path,
+            filesize($path),
+            UPLOAD_ERR_OK,
+            'Foo.txt',
+            'text/plain',
+        );
+
+        self::assertFileDoesNotExist($options['archive']);
+        $result = $filter->__invoke($upload);
```
ramchale, line 140: If you want to avoid the Laminas\Diactoros\UploadedFile import
```suggestion
        $mockStream = $this->createMock(StreamInterface::class);
        $mockStream->expects(self::once())->method('getMetadata')->willReturn($path);

        $mockFileUpload = $this->createMock(UploadedFileInterface::class);
        $mockFileUpload->expects(self::once())->method('getStream')->willReturn($mockStream);

        self::assertFileDoesNotExist($options['archive']);
        $result = $filter->__invoke($mockFileUpload);
```

ramchale (commented):
Looks really good

Added a couple of suggestions that I think mitigate the need to add the dev dependency for laminas/laminas-diactoros, but it depends how much mocking you want to use.

test/File/FileInformationTest.php
froschdesign: The `UploadedFile` is simple and therefore I would not use a mock here.
And it's also much easier to read.

ramchale: That's fair. The mocking is little more complex. laminas/laminas-diactoros is only being imported for two tests, so this is just an alternative approach for if there's more value in not introducing a new dependency, but it is a tradeoff. I'm not precious either way

.gitignore
```diff
@@ -6,3 +6,5 @@
 /laminas-mkdoc-theme/
 /phpunit.xml
 /vendor/
+/documentation-theme/
+.markdownlint.json
```
froschdesign, line 10: Why are these entries needed?

gsteel: These are for local development - If you're checking the docs locally, it'll prevent you from accidentally committing those files. Want a revert?

froschdesign: I have no problem with the entries, I asked because we did not have them before.

froschdesign (approved):
Great improvements!

src/Compress/AggregateArchiveAdapterResolver.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Laminas\Filter\Compress;
+
+use Laminas\Filter\Exception\RuntimeException;
+use Laminas\Filter\File\FileInformation;
+
+use function array_values;
+use function sprintf;
+
+final class AggregateArchiveAdapterResolver implements ArchiveAdapterResolverInterface
```
gsteel, line 13: This "Aggregate Adapter Resolver" can be used to compose a chain of concrete adapter resolvers. The first successful resolver wins.

src/Compress/Bz2Adapter.php
```diff
@@ -0,0 +1,69 @@
...
+use function sprintf;
+
+/**
+ * Compression adapter for Bz2
+ *
+ * @psalm-type Options = array{
+ *     blocksize?: int<1, 9>|null,
+ * }
+ */
+final class Bz2Adapter implements StringCompressionAdapterInterface
```
gsteel, line 24: The BZ2 adapter now only supports string content.

In future, it will be possible to introduce a `FileCompressionInterface` that this adapter _could_ also implement so that we can filter an arbitrary string, or a single un-compressed file to `somefile.bz2`

src/Compress/FileExtensionArchiveAdapterResolver.php
```diff
@@ -0,0 +1,34 @@
...
+
+use function basename;
+use function pathinfo;
+use function sprintf;
+use function str_ends_with;
+use function strtolower;
+
+use const PATHINFO_EXTENSION;
+
+final class FileExtensionArchiveAdapterResolver implements ArchiveAdapterResolverInterface
```
gsteel, line 18: Simply matches `whatever.zip` to use the Zip archive adapter based on filename extension. By default it is composed last in the chain, preferring mime-magic based mime detection.

src/Compress/GzAdapter.php
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Compression adapter for Gzip (ZLib)
+ *
+ * @psalm-type Options = array{
+ *     level?: int<0, 9>|null,
+ *     mode?: 'deflate'|'compress',
+ * }
+ */
+final class GzAdapter implements StringCompressionAdapterInterface
```
gsteel, line 25: As per the BZ2 adapter, the GZ adapter now only supports string content.

In future, it will be possible to introduce a `FileCompressionInterface` that this adapter could also implement so that we can filter an arbitrary string, or a single un-compressed file to `somefile.gz`

src/Compress/MimeTypeArchiveAdapterResolver.php
```diff
@@ -0,0 +1,24 @@
...
+declare(strict_types=1);
+
+namespace Laminas\Filter\Compress;
+
+use Laminas\Filter\Exception\RuntimeException;
+use Laminas\Filter\File\FileInformation;
+
+use function sprintf;
+
+final class MimeTypeArchiveAdapterResolver implements ArchiveAdapterResolverInterface
```
gsteel, line 12: Map the detected mime-type of a file to an archive adapter - the default method of adapter selection

src/File/FileInformation.php
```diff
@@ -0,0 +1,116 @@
...
+use function basename;
+use function file_exists;
+use function finfo_open;
+use function is_array;
+use function is_readable;
+use function is_string;
+
+use const FILEINFO_MIME_TYPE;
+
+final class FileInformation
```
gsteel, line 20: General purpose file information utility. This would have been marked as `@internal`, but we use it to match unknown files to archive adapters in the interface. Marking it as `@internal` would have been a pain in the ass for anyone who wants to implement a different detection strategy

gsteel (commented):
Adding some notes to the PR for posterity!

PR #198, opened 2024-11-11
Refactor RealPath

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Refactor RealPath for #177

### DIFF
--- a/docs/book/v3/standard-filters.md
+++ b/docs/book/v3/standard-filters.md
@@ -1125,8 +1125,7 @@ For more complex usage, read the
 
 ## RealPath
 
-This filter will resolve given links and pathnames, and returns the canonicalized
-absolute pathnames.
+This filter will resolve given links and pathnames, and returns the canonicalized absolute pathnames.
 
 ### Supported Options
 
@@ -1137,13 +1136,12 @@ The following options are supported for `Laminas\Filter\RealPath`:
 
 ### Basic Usage
 
-For any given link or pathname, its absolute path will be returned. References
-to `/./`, `/../` and extra `/` sequences in the input path will be stripped. The
-resulting path will not have any symbolic links, `/./`, or `/../` sequences.
+For any given link or pathname, its absolute path will be returned.
+References to `/./`, `/../` and extra `/` sequences in the input path will be stripped.
+The resulting path will not have any symbolic links, `/./`, or `/../` sequences.
 
-`Laminas\Filter\RealPath` will return `FALSE` on failure, e.g. if the file does not exist. On BSD
-systems `Laminas\Filter\RealPath` doesn't fail if only the last path component doesn't exist, while
-other systems will return `FALSE`.
+`Laminas\Filter\RealPath` will return the value passed to the filter on failure, e.g. if the file does not exist.
+On BSD systems `Laminas\Filter\RealPath` doesn't fail if only the last path component doesn't exist, while other systems will return the value passed to the filter.
 
 ```php
 $filter = new Laminas\Filter\RealPath();
@@ -1155,12 +1153,11 @@ $filtered = $filter->filter($path);
 
 ### Non-Existing Paths
 
-Sometimes it is useful to get paths to files that do n0t exist; e.g., when you
-want to get the real path for a path you want to create. You can then either
-provide a `FALSE` `exists` value at initiation, or use `setExists()` to set it.
+Sometimes it is useful to get paths to files that do not exist; e.g., when you want to get the real path for a path you want to create.
+You can then provide `false` for the `exists` option during construction.
 
 ```php
-$filter = new Laminas\Filter\RealPath(false);
+$filter = new Laminas\Filter\RealPath(['exists' => false]);
 $path = '/www/var/path/../../non/existing/path';
 $filtered = $filter->filter($path);
 
--- a/src/RealPath.php
+++ b/src/RealPath.php
@@ -4,122 +4,58 @@ declare(strict_types=1);
 
 namespace Laminas\Filter;
 
-use Laminas\Stdlib\ErrorHandler;
-use Traversable;
-
 use function array_pop;
 use function explode;
+use function file_exists;
 use function getcwd;
 use function implode;
 use function is_string;
-use function preg_match;
-use function preg_replace;
 use function realpath;
 use function str_starts_with;
-use function stripos;
-use function substr;
 
 use const DIRECTORY_SEPARATOR;
-use const PHP_OS;
 
 /**
  * @psalm-type Options = array{
- *     exists: bool,
- *     ...
+ *     exists?: bool,
  * }
- * @template TOptions of Options
- * @extends AbstractFilter<TOptions>
+ * @implements FilterInterface<string>
  */
-final class RealPath extends AbstractFilter
+final class RealPath implements FilterInterface
 {
-    /** @var TOptions $options */
-    protected $options = [
-        'exists' => true,
-    ];
+    private readonly bool $pathMustExist;
 
-    /**
-     * @param  bool|Traversable|Options $existsOrOptions Options to set
-     */
-    public function __construct($existsOrOptions = true)
+    /** @param Options $options */
+    public function __construct(array $options = [])
     {
-        if ($existsOrOptions !== null) {
-            if (! static::isOptions($existsOrOptions)) {
-                $this->setExists($existsOrOptions);
-            } else {
-                $this->setOptions($existsOrOptions);
-            }
-        }
+        $this->pathMustExist = $options['exists'] ?? true;
     }
 
-    /**
-     * Sets if the path has to exist
-     * TRUE when the path must exist
-     * FALSE when not existing paths can be given
-     *
-     * @param  bool $flag Path must exist
-     * @return self
-     */
-    public function setExists($flag = true)
-    {
-        $this->options['exists'] = (bool) $flag;
-        return $this;
-    }
-
-    /**
-     * Returns true if the filtered path must exist
-     *
-     * @return bool
-     */
-    public function getExists()
-    {
-        return $this->options['exists'];
-    }
-
-    /**
-     * Defined by Laminas\Filter\FilterInterface
-     *
-     * Returns realpath($value)
-     *
-     * If the value provided is non-scalar, the value will remain unfiltered
-     *
-     * @psalm-return ($value is string ? string : mixed)
-     */
     public function filter(mixed $value): mixed
     {
         if (! is_string($value)) {
             return $value;
         }
-        $path = (string) $value;
 
-        if ($this->options['exists']) {
-            return realpath($path);
+        if ($this->pathMustExist && ! file_exists($value)) {
+            return $value;
         }
 
-        ErrorHandler::start();
-        $realpath = realpath($path);
-        ErrorHandler::stop();
-        if ($realpath !== false) {
-            return $realpath;
+        $realPath = realpath($value);
+
+        if ($realPath !== false) {
+            return $realPath;
         }
 
-        $drive = '';
-        if (stripos(PHP_OS, 'WIN') === 0) {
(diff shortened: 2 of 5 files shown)


docs/book/v3/standard-filters.md
```diff
@@ -1156,11 +1156,11 @@ $filtered = $filter->filter($path);
 ### Non-Existing Paths
 
 Sometimes it is useful to get paths to files that do n0t exist; e.g., when you
```
```suggestion
Sometimes it is useful to get paths to files that do not exist; e.g., when you
```

docs/book/v3/standard-filters.md
```diff
@@ -1156,11 +1156,11 @@ $filtered = $filter->filter($path);
...
-want to get the real path for a path you want to create. You can then either
-provide a `FALSE` `exists` value at initiation, or use `setExists()` to set it.
+want to get the real path for a path you want to create. You can then
+provide a `FALSE` `exists` option at initiation.
```
```suggestion
provide `false` for the `exists` option during construction.
```

gsteel: Also, whilst the current docs don't conform to this rule - can you please re-format _(just the changes)_ to one sentence per line please. This is a general, un-documented docs style preference, that hopefully we'll add to a style guide one day 😬

src/RealPath.php
```diff
@@ -24,74 +23,28 @@
...
     {
         if (! is_string($value)) {
             return $value;
         }
         $path = (string) $value;
 
-        if ($this->options['exists']) {
+        if ($this->pathMustExist) {
             return realpath($path);
         }
```
gsteel, line 43: Can we get rid of `ErrorHandler::start()` and `stop()` here? I'm not convinced that any warnings/errors are emitted: https://3v4l.org/jT1XV

src/RealPath.php
```diff
@@ -24,74 +23,28 @@
...
-     * If the value provided is non-scalar, the value will remain unfiltered
-     *
-     * @psalm-return ($value is string ? string : mixed)
-     */
     public function filter(mixed $value): mixed
     {
         if (! is_string($value)) {
             return $value;
         }
         $path = (string) $value;
```
gsteel: This cast is not necessary

src/RealPath.php
```diff
@@ -24,74 +23,28 @@
...
 
-        if ($this->options['exists']) {
+        if ($this->pathMustExist) {
             return realpath($path);
```
gsteel: Typical filter behaviour should be to either return the successfully filtered value or the un-filtered value. This is potentially returning `false`, so I think there should be conditional to return `$value` when `realpath() === false`.

```php
// file_exists will return appropriate result for args such as '/some/../other/../foo/bar'
if ($this->pathMustExist && ! file_exists($value)) {
    return $value;
}

if ($this->pathMustExist) {
    return realpath($value);
}
```

This behaviour change will need a sentence in the migration docs and the tests will need changing too.

src/RealPath.php
```diff
@@ -24,74 +23,28 @@
 
 /**
  * @psalm-type Options = array{
- *     exists: bool,
- *     ...
+ *     exists?: bool,
  * }
- * @template TOptions of Options
- * @extends AbstractFilter<TOptions>
+ * @implements FilterInterface<string|array<array-key, string|mixed>>
```
gsteel:
```suggestion
 * @implements FilterInterface<string>
```

The successfully filtered value is a string - therefore the return type of `filter()` will be `mixed|string` - we might be able to narrow this to `non-empty-string` further if its not a massive pain in the ass.

src/RealPath.php
gsteel: Also for all the code below, it remains untested because CI does not run windows 🤔

gsteel (changes requested):
Thanks @ramchale - The windows related shenanigans need some more attention… I'll have a think about running CI on windows, or perhaps also discuss on slack general windows support…

docs/book/v3/standard-filters.md
froschdesign: > # Writing Style
> 
> ## One Sentence per-line Method
> 
> Documentation should be written using [the one sentence per-line method](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line). 
> There are a number of excellent reasons for writing in this style, including:
> 
> - It prevents reflows (meaning a change early in the paragraph won’t cause the remaining lines in the paragraph to reposition)
> - You can easily swap/move sentences
> - You can easily separate or join paragraphs
> - You can comment out sentences or add commentary to them
> - You can spot sentences which are too long or sentences that vary widely in length
> - You can spot redundant (and thus mundane) patterns in your writing
> 
> Take the following content as an example.
> 
> ```markdown
> The concrete `TableGateway` object uses constructor injection for getting
> dependencies and options into the instance. The table name and an instance of
> an `Adapter` are all that is required to create an instance. Out of the box, this 
> implementation makes no assumptions about table structure or metadata, and 
> when `select()` is executed, a simple `ResultSet` object with the populated 
> `Adapter`'s `Result` (the datasource) will be returned and ready for iteration.
> ```
> 
> While it can be easily read in any editor, as it does not require the reader to scroll, that simplicity comes at the cost of maintainability. 
> For example, if you needed to put the second sentence before the first, or move it to any other location, it would need to be cut out of the surrounding paragraph before it could be moved.
> 
> Compare that with the following sample content.
> 
> ```markdown
> The concrete `TableGateway` object uses constructor injection for getting dependencies and options into the instance. 
> The table name and an instance of an `Adapter` are all that is required to create an instance.
> Out of the box, this implementation makes no assumptions about table structure or metadata, and when `select()` is executed, a simple `ResultSet` object with the populated `Adapter`'s `Result` (the datasource) will be returned and ready for iteration.
> ```
> 
> Yes, the content, by default scrolls out of view. 
> However, most text editors and IDEs support "_soft wrapping_", making this a non-issue. 
> Using this approach, the sentence doesn't need to be extracted from the surrounding paragraph before it can be moved.

gsteel: Sorry Frank! the _documented_ style guide… Where is it?

ramchale: Will do. Thanks both

src/RealPath.php
ramchale: > Can we get rid of `ErrorHandler::start()` and `stop()` here? I'm not convinced that any warnings/errors are emitted: https://3v4l.org/jT1XV

Looks like we can. I played around with the open_baseline restriction just to make sure it wasn't something to do with that, but it also just returns false if outside of the allowed directories.

src/RealPath.php
ramchale: I think there's a strong possibility of some people relying on this returning false for validation (which I know it isn't intended for 😄). As the behavior for passing in a path that doesn't exist would change to returning the same path back, with no indication as to whether that's because it matches the realpath, is it okay to suggest introducing Laminas\Validator\File\Exists in the docs if people aren't using it?

gsteel: Yes! Improving docs to state that filters shouldn't be used for validation is a worthy improvement! As filter does not depend on validator, I don't think its _necessary_ to mention laminas-validator itself, but write whatever makes most sense to the user - Typically filtering happens prior to validation, so returning the un-modified path argument when it is un-filterable is good, because it should be validated by `..\File\Exists` afterwards.

gsteel:
Further to the Windows specific code here - can you please remove it, and any related tests?

We should also document dropping windows support in this filter.

src/RealPath.php
ramchale: Cool. Thanks

docs/book/v3/standard-filters.md
```diff
@@ -1137,13 +1136,12 @@ The following options are supported for `Laminas\Filter\RealPath`:
...
-to `/./`, `/../` and extra `/` sequences in the input path will be stripped. The
-resulting path will not have any symbolic links, `/./`, or `/../` sequences.
+For any given link or pathname, its absolute path will be returned.
+References to `/./`, `/../` and extra `/` sequences in the input path will be stripped.
+The resulting path will not have any symbolic links, `/./`, or `/../` sequences.
 
-`Laminas\Filter\RealPath` will return `FALSE` on failure, e.g. if the file does not exist. On BSD
-systems `Laminas\Filter\RealPath` doesn't fail if only the last path component doesn't exist, while
-other systems will return `FALSE`.
+`Laminas\Filter\RealPath` will return `FALSE` on failure, e.g. if the file does not exist.
```
```suggestion
`Laminas\Filter\RealPath` will return the un-filtered input on failure, for example, if the input is not a string, the value cannot be interpreted as a file path, or, the path does not exist when existence is a configured requirement.
```

docs/book/v3/standard-filters.md
```diff
@@ -1137,13 +1136,12 @@ The following options are supported for `Laminas\Filter\RealPath`:
...
+On BSD systems `Laminas\Filter\RealPath` doesn't fail if only the last path component doesn't exist, while other systems will return `false`.
```
gsteel: Can we verify this? also, it shouldn't return false.

src/RealPath.php
```diff
@@ -42,48 +29,18 @@ public function filter(mixed $value): mixed
...
-                [, $drive, $path] = $matches;
-            } else {
-                $cwd   = getcwd();
-                $drive = substr($cwd, 0, 2);
-                if (! str_starts_with($path, DIRECTORY_SEPARATOR)) {
-                    $path = substr($cwd, 3) . DIRECTORY_SEPARATOR . $path;
-                }
-            }
-        } elseif (! str_starts_with($path, DIRECTORY_SEPARATOR)) {
-            $path = getcwd() . DIRECTORY_SEPARATOR . $path;
```
gsteel, line 118: Conditionally preprending `getcwd()` is a feature we should probably keep and test:

```php
$filter->filter('foo.txt'); // "/var/www/project/foo.txt"
```

src/RealPath.php
```diff
@@ -42,48 +29,18 @@ public function filter(mixed $value): mixed
...
-        $stack = [];
-        $parts = explode(DIRECTORY_SEPARATOR, $path);
-        foreach ($parts as $dir) {
-            if ($dir !== '' && $dir !== '.') {
-                if ($dir === '..') {
-                    array_pop($stack);
-                } else {
-                    $stack[] = $dir;
-                }
-            }
```
gsteel: Dropping this code means that you cannot fabricate the `realpath` for a non-existent file, i.e. `realpath('/not-there') === false`

gsteel (changes requested):
Thanks @ramchale, but, we still need to "make up" the real path for non-existent files when appropriate

ramchale:
> Thanks @ramchale, but, we still need to "make up" the real path for non-existent files when appropriate

🤦 sorry. Not my best work. I've reworked some of the tests to better cover that with the return value changes.
I've added a test that should check whether the statement about BSD is true, but don't have access to a BSD machine to check and the docker support attempts seem to have petered out.

gsteel (changes requested):
Thank you @ramchale - this is looking _much_ better - could we improve the test `testExistingFileReturnsRealPath` with a data provider along the lines of

```php
[
   __DIR__ . '/_files/foo/../bar/../file.1',
   __DIR__ . '/_files/././file.1',
   __DIR__ . '///_files///file.1',
   // etc…
]
```

ramchale:
Thanks @gsteel . I've added those.
`__DIR__ . '/_files/foo/../bar/../file.1'` is invalid, but seemed like the sort of edge-case that might come up, so I've dropped it in it's own test.
Once you're happy with the changes, is there any issue with me squashing this before it's merged to cut down on some of the undoing of changes in the commit history?

gsteel (approved):
Thanks again @ramchale - Great improvement

Let me know when you're done squashing and I'll merge 👍

ramchale:
@gsteel Thanks. Should be ready now

gsteel (approved):
Thanks!

PR #200, opened 2024-11-18
Refactor `MonthSelect` removing inheritance and improving tests

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | no

### Description

Draft pull request to discuss desired behavior of MonthSelect in v3. Some of this will also be relevant to other classes extending AbstractDateDropdown.

* Currently this throws a RuntimeException if an array is passed to the filter with the incorrect number of date parts. Should just pass through unprocessed for consistency with other filters?
* How do you want this to handle months that are outside of a valid date range?
* How do you want to handle values for month that are not numerical (strings, bools, null, etc)? (I'm guessing the original value should be returned unprocessed, but better not to assume)

### DIFF
--- a/src/MonthSelect.php
+++ b/src/MonthSelect.php
@@ -4,24 +4,88 @@ declare(strict_types=1);
 
 namespace Laminas\Filter;
 
+use function filter_var;
+use function is_array;
+use function is_numeric;
+use function sprintf;
+
+use const FILTER_VALIDATE_INT;
+
 /**
  * @psalm-type Options = array{
  *     null_on_empty?: bool,
  *     null_on_all_empty?: bool,
- *     ...
  * }
- * @psalm-type InputArray = array{
- *       year: numeric,
- *       month: numeric,
- *  }
- * @template TOptions of Options
- * @template-extends AbstractDateDropdown<TOptions, InputArray>
+ * @implements FilterInterface<string|null>
  */
-final class MonthSelect extends AbstractDateDropdown
+final class MonthSelect implements FilterInterface
 {
+    private readonly bool $returnNullIfAnyFieldEmpty;
+    private readonly bool $returnNullIfAllFieldsEmpty;
+
+    /** @param Options $options */
+    public function __construct(array $options = [])
+    {
+        $this->returnNullIfAnyFieldEmpty  = $options['null_on_empty'] ?? false;
+        $this->returnNullIfAllFieldsEmpty = $options['null_on_all_empty'] ?? false;
+    }
+
+    public function __invoke(mixed $value): mixed
+    {
+        return $this->filter($value);
+    }
+
     /**
-     * Year-Month
+     * Returns the result of filtering $value
+     *
+     * @template T
+     * @param T $value
+     * @return string|null|T
      */
-    protected string $format      = '%2$s-%1$s';
-    protected int $expectedInputs = 2;
+    public function filter(mixed $value): mixed
+    {
+        if (! is_array($value)) {
+            return $value;
+        }
+
+        $month = $value['month'] ?? null;
+        /** @var mixed $month */
+        $month = $month === '' ? null : $month;
+
+        $year = $value['year'] ?? null;
+        /** @var mixed $year */
+        $year = $year === '' ? null : $year;
+
+        if ($this->returnNullIfAnyFieldEmpty && ($month === null || $year === null)) {
+            return null;
+        }
+
+        if ($this->returnNullIfAllFieldsEmpty && $month === null && $year === null) {
+            return null;
+        }
+
+        if (! $this->isParsableAsDateValue($month, 1, 12) || ! $this->isParsableAsDateValue($year, 0, 9999)) {
+            /** @psalm-var T */
+            return $value;
+        }
+
+        return sprintf('%d-%02d', $year, $month);
+    }
+
+    /** @psalm-assert-if-true int $value */
+    private function isParsableAsDateValue(mixed $value, int $lowestValue, int $highestValue): bool
+    {
+        if (
+            ! is_numeric($value)
+            || filter_var(
+                $value,
+                FILTER_VALIDATE_INT,
+                ['options' => ['min_range' => $lowestValue, 'max_range' => $highestValue]]
+            ) === false
+        ) {
+            return false;
+        }
+
+        return true;
+    }
 }
--- a/test/MonthSelectTest.php
+++ b/test/MonthSelectTest.php
@@ -4,37 +4,65 @@ declare(strict_types=1);
 
 namespace LaminasTest\Filter;
 
-use Laminas\Filter\Exception\RuntimeException;
 use Laminas\Filter\MonthSelect as MonthSelectFilter;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 
+/** @psalm-import-type Options from MonthSelectFilter */
 class MonthSelectTest extends TestCase
 {
+    /** @param Options $options */
     #[DataProvider('provideFilter')]
     public function testFilter(array $options, array $input, ?string $expected): void
     {
-        $sut = new MonthSelectFilter();
-        $sut->setOptions($options);
+        $sut = new MonthSelectFilter($options);
+
         self::assertSame($expected, $sut->filter($input));
     }
 
-    /** @return list<array{0: array, 1: array, 2: string|null}> */
+    /** @return list<array{0: Options, 1: array, 2: string|null}> */
     public static function provideFilter(): array
     {
         return [
+            [[], ['year' => '2014', 'month' => '2'], '2014-02'],
             [[], ['year' => '2014', 'month' => '10'], '2014-10'],
-            [['nullOnEmpty' => true], ['year' => null, 'month' => '10'], null],
+            [[], ['year' => 2014, 'month' => 10], '2014-10'],
             [['null_on_empty' => true], ['year' => null, 'month' => '10'], null],
-            [['nullOnAllEmpty' => true], ['year' => null, 'month' => null], null],
+            [['null_on_empty' => true], ['month' => null], null],
+            [['null_on_empty' => true], ['year' => null], null],
             [['null_on_all_empty' => true], ['year' => null, 'month' => null], null],
+            [['null_on_all_empty' => true], [], null],
+            [['null_on_all_empty' => true], ['year' => '', 'month' => ''], null],
         ];
     }
 
-    public function testInvalidInput(): void
+    #[DataProvider('provideInvalidFilterValues')]
+    public function testInvalidInput(mixed $value): void
     {
-        $this->expectException(RuntimeException::class);
         $sut = new MonthSelectFilter();
-        $sut->filter(['year' => '2120']);
+
+        self::assertSame($value, $sut->filter($value));
+    }
+
+    /** @return array<string, array{0: mixed}> */
+    public static function provideInvalidFilterValues(): array
+    {
+        return [
+            'empty array'           => [[]],
+            'missing year'          => [['month' => '10']],
+            'missing month'         => [['year' => '2023']],
+            'passed bool'           => [true],
+            'passed string'         => ['string'],
+            'passed int'            => [10],
+            'passed float'          => [10.5],
+            'invalid keys'          => [['should be year' => '2014', 'should be month' => '10']],
+            'year is invalid type'  => [['year' => true, 'month' => '09']],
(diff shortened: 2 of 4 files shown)


gsteel:
Looks to me like expected input is, like you say, an associative array of varying lengths depending on the descendant class that should produce a formatted string. The filter requires that array members be specificly `year`, `month` etc, so it's already very restrictive - I really don't see the point of this filter at all - I mean, on the frontend, why not just use an `<input type="date|datetime-local">` - this filter pre-dates HTML5, so I'm wondering if we should just deprecate all of it and remove it completely.

cc @froschdesign

ramchale:
Cool. Does that apply to the other two that inherit from AbstractDateDropdown (DateSelect and DateTimeSelect), as they share a lot of the same behavior at the moment?

gsteel:
I'd say all 4 classes, MonthSelect, Abstract, DateSelect & DateTimeSelect - Let's wait for @froschdesign to give us an opinion too 👍

froschdesign:
> I'd say all 4 classes, MonthSelect, Abstract, DateSelect & DateTimeSelect…

Unfortunately, that won't work, because laminas-form uses these filters:

- `Laminas\Form\Element\DateSelect`: https://github.com/laminas/laminas-form/blob/653c869d10c361027ae6c660c991ec3e3f38ed65/src/Element/DateSelect.php#L175-L193
- `Laminas\Form\Element\DateTimeSelect`: https://github.com/laminas/laminas-form/blob/653c869d10c361027ae6c660c991ec3e3f38ed65/src/Element/DateTimeSelect.php#L335-L353
- `Laminas\Form\Element\MonthSelect`: https://github.com/laminas/laminas-form/blob/653c869d10c361027ae6c660c991ec3e3f38ed65/src/Element/MonthSelect.php#L325-L343

ramchale:
> > I'd say all 4 classes, MonthSelect, Abstract, DateSelect & DateTimeSelect…
> 
> Unfortunately, that won't work, because laminas-form uses these filters:
> 
>     * `Laminas\Form\Element\DateSelect`: https://github.com/laminas/laminas-form/blob/653c869d10c361027ae6c660c991ec3e3f38ed65/src/Element/DateSelect.php#L175-L193
> 
>     * `Laminas\Form\Element\DateTimeSelect`: https://github.com/laminas/laminas-form/blob/653c869d10c361027ae6c660c991ec3e3f38ed65/src/Element/DateTimeSelect.php#L335-L353
> 
>     * `Laminas\Form\Element\MonthSelect`: https://github.com/laminas/laminas-form/blob/653c869d10c361027ae6c660c991ec3e3f38ed65/src/Element/MonthSelect.php#L325-L343

Thanks both. Does that bring us back to the questions about behavior in v3 or is there more to consider?

gsteel:
OK, so I think that in all erroneous cases, the filter(s) should return the unprocessed input that will hopefully then fail subsequent validation - so, out of range, incorrect array shapes etc should not throw exceptions, but just silently pass through unchanged.

I think we have to work on the premise that we're working with data that hasn't been validated. In typical usage with `input-filter`, throwing an exception for malformed input prevents validators from doing their job and reporting to the user that the input was rejected.

froschdesign:
> OK, so I think that in all erroneous cases, the filter(s) should return the unprocessed input that will hopefully then fail subsequent validation - so, out of range, incorrect array shapes etc should not throw exceptions, but just silently pass through unchanged.

Based on the form elements from above: The validators `Date` and `Regex` can handle the unprocessed input arrays so that the validators can fulfil their jobs in all cases and produce the related error messages.

ramchale:
I think this covers the desired behaviour. Let me know if there's anything not quite right, or any other test cases that would be useful. Once we've got the code doing what it should I'll look at adding some docs.

Also, psalm isn't loving returning $value after `$month = $value['month'] ?? null;`. While I don't want to restructure just to appease static analysis, if you've got any suggestions to avoid the error suppression that'd be useful.

src/MonthSelect.php
```diff
@@ -4,24 +4,84 @@
...
  * }
  * @psalm-type InputArray = array{
- *       year: numeric,
- *       month: numeric,
+ *       year: int,
+ *       month: int,
  *  }
- * @template TOptions of Options
- * @template-extends AbstractDateDropdown<TOptions, InputArray>
+ * @template TInput of array<array-key, numeric>
```
gsteel: `TInput` can be deleted - I don't think it's usead anymore

src/MonthSelect.php
```diff
@@ -4,24 +4,84 @@
...
  *     null_on_empty?: bool,
  *     null_on_all_empty?: bool,
- *     ...
  * }
  * @psalm-type InputArray = array{
- *       year: numeric,
- *       month: numeric,
+ *       year: int,
+ *       month: int,
  *  }
```
gsteel: `InputArray` can be delete I think

src/MonthSelect.php
```diff
@@ -4,24 +4,84 @@
...
+        $this->returnNullIfAnyFieldEmpty  = $options['null_on_empty'] ?? false;
+        $this->returnNullIfAllFieldsEmpty = $options['null_on_all_empty'] ?? false;
+    }
+
+    public function __invoke(mixed $value): mixed
+    {
+        return $this->filter($value);
+    }
+
+    /** @psalm-suppress InvalidReturnType */
```
gsteel: Not sure why this suppression is necessary, possibly, we should make `@implements FilterInterface<string|null>` instead of just `string`

src/MonthSelect.php
```diff
@@ -4,24 +4,84 @@
...
+        if ($this->returnNullIfAnyFieldEmpty && ($month === null || $year === null)) {
+            return null;
+        }
+
+        if ($this->returnNullIfAllFieldsEmpty && $month === null && $year === null) {
+            return null;
+        }
+
+        if (! $this->isParsableAsDateValue($month, 1, 12) || ! $this->isParsableAsDateValue($year, 0, 9999)) {
+            /** @psalm-suppress InvalidReturnStatement */
```
gsteel: Hopefully, with `@implements` updated, we can drop these suppressions too.

src/MonthSelect.php
```diff
@@ -4,24 +4,84 @@
...
+            return $value;
+        }
+
+        /** @psalm-suppress MixedArgument */
+        return sprintf('%s-%s', $year, $month);
+    }
+
+    private function isParsableAsDateValue(mixed $value, int $lowestValue, int $highestValue): bool
```
gsteel, line 76: I was hoping we could annotate this with `@psalm-assert-if-true int<T1, T2> $value` but it's not possible: https://psalm.dev/r/4a9988cce3

Alternatively, you can annotate as `@psalm-assert-if-true int $value`, that will fix the MixedArgument issue in `filter` I'd expect.

src/MonthSelect.php
```diff
@@ -4,24 +4,84 @@
...
+            return null;
+        }
+
+        if (! $this->isParsableAsDateValue($month, 1, 12) || ! $this->isParsableAsDateValue($year, 0, 9999)) {
+            /** @psalm-suppress InvalidReturnStatement */
+            return $value;
+        }
+
+        /** @psalm-suppress MixedArgument */
+        return sprintf('%s-%s', $year, $month);
```
gsteel: Should probably be `%d-%02d` so that months are zero padded, as I expect the output would be used to form an ISO date string (??)

test/MonthSelectTest.php
```diff
@@ -24,17 +23,41 @@ public static function provideFilter(): array
...
+    public function testInvalidInput(mixed $value): void
     {
-        $this->expectException(RuntimeException::class);
         $sut = new MonthSelectFilter();
-        $sut->filter(['year' => '2120']);
+
+        self::assertSame($value, $sut->filter($value));
+    }
+
+    public static function provideInvalidFilterValues(): array
```
gsteel, line 48:
```suggestion
    /** @return array<string, array{0: mixed}> */
    public static function provideInvalidFilterValues(): array
```

test/MonthSelectTest.php
```diff
@@ -14,8 +13,8 @@ class MonthSelectTest extends TestCase
     #[DataProvider('provideFilter')]
     public function testFilter(array $options, array $input, ?string $expected): void
     {
-        $sut = new MonthSelectFilter();
-        $sut->setOptions($options);
+        /** @psalm-suppress MixedArgumentTypeCoercion */
```
gsteel: If you `@psalm-import-type Options from MonthSelect` on the class, add `@param Options $options` here and `@return list<array{0: Options, 1: mixed, 2: string|null}>` on the data provider, no inline suppressions will be needed. As a general rule, only ever add inline supressions to code when you are knowingly providing an invalid argument to force an exception or something in a test. Issues belong in the baseline when they are not getting fixed 👍

gsteel (changes requested):

ramchale:
Thanks. Still needs the suppression for the return statement, but the rest is looking a lot healthier

src/MonthSelect.php
```diff
@@ -4,24 +4,78 @@
...
+        $this->returnNullIfAnyFieldEmpty  = $options['null_on_empty'] ?? false;
+        $this->returnNullIfAllFieldsEmpty = $options['null_on_all_empty'] ?? false;
+    }
+
+    public function __invoke(mixed $value): mixed
+    {
+        return $this->filter($value);
+    }
+
+    public function filter(mixed $value): mixed
```
gsteel, line 45:
```suggestion
    /**
     * Returns the result of filtering $value
     *
     * @template T
     * @param T $value
     * @return string|null|T
     */
    public function filter(mixed $value): mixed
```

Turns out that re-implementing the template is necessary here so that we can force `$value as T` later on…

src/MonthSelect.php
```diff
@@ -4,24 +4,78 @@
...
+        if ($this->returnNullIfAnyFieldEmpty && ($month === null || $year === null)) {
+            return null;
+        }
+
+        if ($this->returnNullIfAllFieldsEmpty && $month === null && $year === null) {
+            return null;
+        }
+
+        if (! $this->isParsableAsDateValue($month, 1, 12) || ! $this->isParsableAsDateValue($year, 0, 9999)) {
+            /** @psalm-suppress InvalidReturnStatement */
```
gsteel:
```suggestion
            /** @psalm-var T */
```

This fixes the more specific inference of `$value` already done by Psalm - because the value is made more specific by the code, we're reducing it to `T` _(the input)_ instead of `array{month?:mixed, year?: mixed, ...<array-key, mixed>}` which we'd have to add to `@implements FilterInterface<whatever>` which is not what we want.

src/MonthSelect.php
```diff
@@ -4,24 +4,78 @@
...
+    public function filter(mixed $value): mixed
+    {
+        if (! is_array($value)) {
+            return $value;
+        }
+
+        /** @var mixed $month */
+        $month = $value['month'] ?? null;
+        /** @var mixed $year */
+        $year = $value['year'] ?? null;
```
gsteel:
```suggestion
        $month = $value['month'] ?? null;
        /** @var mixed $month */
        $month = $month === '' ? null : $month;
        $year  = $value['year'] ?? null;
        /** @var mixed $year */
        $year = $year === '' ? null : $year;
```
Sorry to be a pain, but we should probably coalesce the empty string to null for both `$year` and `$month` before comparison too, and add those to the tests because if you've set these options to true, you probably want `null` when `$input === ['month' => '', 'year' => '']`

gsteel (changes requested):
Thanks @ramchale - I figured out the reason for the `InvalidReturnStatement` - Sorry to be pedantic 😬

ramchale:
> Thanks @ramchale - I figured out the reason for the `InvalidReturnStatement` - Sorry to be pedantic 😬

No. Really appreciate the feedback and help. Thanks

gsteel (approved):
Thanks @ramchale 👍

gsteel:
Is there anything else needed here or can I merge it?

ramchale:
> Is there anything else needed here or can I merge it?

I've got some accompanying documentation I can tidy up and add to the v2_to_v3.md and standard-filters.md this afternoon. So might want to hold off for that?

gsteel:
> I've got some accompanying documentation I can tidy up and add to the v2_to_v3.md and standard-filters.md this afternoon. So might want to hold off for that?

Cool - I'll wait for the docs so they go in together 👍

gsteel (approved):
Thanks again @ramchale 👌

PR #202, opened 2024-12-04
Remove sections on Alnum, Alpha, NumberFormat, and NumberParse

froschdesign:
This commit removes detailed documentation for the Alnum, Alpha, NumberFormat, and NumberParse filters from the standard-filters documentation. The removed sections included the filters' descriptions, supported options, basic usage examples, and notes on supported languages. This helps in streamlining the document and possibly moving these details to more specific locations or documents dedicated to those filters.

See: https://docs.laminas.dev/laminas-i18n/filters/introduction/


froschdesign:
I think I can find later another example for the "Inflector" section, then we can simplify the part and laminas-i18n does not have to be installed.

gsteel (approved):

PR #204, opened 2024-12-04
Deprecate `FilterProviderInterface`

gsteel:
This interface is MVC Module Manager specific and does not belong in this library


src/FilterProviderInterface.php
```diff
@@ -7,6 +7,8 @@
 /**
  * Implement this interface within Module classes to indicate that your module
  * provides filter configuration for the FilterPluginManager.
+ *
+ * @deprecated Since 2.40.0 This interface will be removed in version 3.0 without replacement
```
froschdesign, line 11: Must also be listed in the migration guide.

gsteel: Handled in #205

PR #205, opened 2024-12-04
Remove `FilterProviderInterface`

gsteel:
This interface is MVC Module Manager specific and does not belong in this library

See #204


froschdesign:
The documentation must also be removed:

https://github.com/laminas/laminas-filter/blob/08e0bb9e98aa632a0aac51c27d2624a97ccf7261/docs/book/v3/application-integration/providing-filters-via-modules.md?plain=1#L1-L16

gsteel:
> The documentation must also be removed:

I've retained and updated the docs to point to the same interface in the Module Manager lib/docs

src/FilterProviderInterface.php
```diff
@@ -1,23 +0,0 @@
...
-
-declare(strict_types=1);
-
-namespace Laminas\Filter;
-
-use Laminas\ServiceManager\ServiceManager;
-
-/**
- * Implement this interface within Module classes to indicate that your module
- * provides filter configuration for the FilterPluginManager.
```
froschdesign, line 11: But doesn't this also apply to a config provider? What I mean by that is, wouldn't the interface also help the user in a config provider?

gsteel: I don't see how this would help with a Config Provider - MVC doesn't use them and typically they are aggregated en-masse via `__invoke` so the interface would be superfluous.

The only important part I can see for a `ConfigProvider` is knowing the top-level config key `filters` which is covered in the docs

froschdesign: The interface does not really help, because the `__invoke` method is relevant, not the getter method. So kill the interface.

froschdesign (approved):

PR #208, opened 2024-12-10
Refactor `FilterChain`

gsteel:
Refactors filter chain to require the plugin manager as a constructor dependency and…

- Drops `setOptions`
- Removes inheritance
- Implement newly added `FilterChainInterface`
- Remove the constant `DEFAULT_PRIORITY`
- Drops getter and setter for the plugin manager
- Drops the `plugin` method
- Drops `getFilter`
- Stops pretending that chains can be serialized

Introduces `ImmutableFilterChain` which is effectively the same as `FilterChain` but immutable and lacks the `merge` method which has limited use cases.

 `ImmutableFilterChain` and `FilterChainInterface` can be backported to 2.x so that libs can switch to it (input-filter, form etc) without breaking BC _(Possibly)_

Docs are still todo but feedback appreciated 👍


froschdesign (approved):
I like deleting code, therefore: great improvement! 👍🏻

gsteel:
@froschdesign 
I've updated the docs for FilterChain and writing filters, could you please cast your eye over them? Thanks :)

froschdesign (approved):
Great! 👍🏻

PR #212, opened 2024-12-21
Refactor RenameTest

ramchale:
- Only remove temp directory after all tests have run
- Use DataProvider for filter config and input variations

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Draft PR to discuss refactoring of RenameTest.

Started looking at this as part of the moving the Rename filter to v3, but this feels like a big enough piece of refactoring in itself to be worth sense checking.

As DataProviders run before any of the test hooks, the creation of the tmp test file is moved to a static method that creates it on first use, and then it's not removed until all the tests in RenameTest have run. Likewise for the sub directory needed for testing moving a file to a target directory.

The cleanup for the indvidual files created by tests has been moved to each test. Could potentially be moved back to tearDown if we're not concerned about the number of redundant calls, and depending on what seems more readable.

### DIFF
--- a/test/File/RenameTest.php
+++ b/test/File/RenameTest.php
@@ -11,9 +11,8 @@ use PHPUnit\Framework\TestCase;
 use stdClass;
 
 use function copy;
-use function dirname;
 use function file_exists;
-use function is_dir;
+use function is_array;
 use function mkdir;
 use function preg_quote;
 use function rmdir;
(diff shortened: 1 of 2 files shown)


test/File/RenameTest.php
```diff
@@ -26,393 +25,328 @@
...
+                    'randomize' => false,
+                ],
+                'expectedFilterResult'  => $oldFile,
+            ],
+        ];
+    }
+
+    #[DataProvider('returnValidFilterInputProvider')]
+    #[DataProvider('returnInvalidFilterInputProvider')]
+    /** @param FilterInput $expectedFilterResult **/
```
gsteel: The `FilterInput` parameter type isn't imported anywhere AFAICT

gsteel (changes requested):
Thanks @ramchale
This test was a real mess huh? LGTM, other than one item.
Hopefully we've got enough in the current 2.x release branch to tag and merge this up into 3.x when its ready 👍

test/File/RenameTest.php
ramchale: Thanks

ramchale:
> Thanks @ramchale This test was a real mess huh? LGTM, other than one item. Hopefully we've got enough in the current 2.x release branch to tag and merge this up into 3.x when its ready 👍

Happy New Year!
Unit testing anything involving the file system is always a pain 😓
Thanks. That should be updated now

gsteel (approved):
Thank you @ramchale :)

PR #214, opened 2025-01-02
Refactor StripTags filter to implement FlterInterface

marcelthole:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | no

### Description
Move StripTags filter also to the implementation via FilterInterface and removed the AbstractFilter


gsteel (approved):
Thanks @marcelthole

I added another commit to further simplify the options as readonly properties without a private setter method. Also, the options were tightened up so that tags are always `array<string, list<string>>` and global attributes are `list<string>`.

I would have done this via review, but was having trouble communicating my intentions!

froschdesign (approved):

PR #215, opened 2025-01-02
Refactor htmlentities filter

marcelthole:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | no

### Description
Move HtmlEntities filter also to the implementation via FilterInterface and removed the AbstractFilter

### DIFF
--- a/docs/book/v3/standard-filters.md
+++ b/docs/book/v3/standard-filters.md
@@ -635,23 +635,16 @@ entity equivalents when possible.
 The following options are supported for `Laminas\Filter\HtmlEntities`:
 
 - `quotestyle`: Equivalent to the PHP `htmlentities()` native function parameter
-  `quote_style`.  This allows you to define what will be done with 'single' and
+  `flags`.  This allows you to define what will be done with 'single' and
   "double" quotes. The following constants are accepted: `ENT_COMPAT`,
   `ENT_QUOTES`, and `ENT_NOQUOTES`, with the default being `ENT_COMPAT`.
-- `charset`: Equivalent to the PHP `htmlentities()` native function parameter
-  `charset`. This defines the character set to be used in filtering. Unlike the
+- `encoding`: Equivalent to the PHP `htmlentities()` native function parameter
+  `encoding`. This defines the character set to be used in filtering. Unlike the
   PHP native function, the default is 'UTF-8'. See the [PHP htmlentities
   manual](http://php.net/htmlentities) for a list of supported character sets.
-
-  This option can also be set via the `$options` parameter as a Traversable
-  object or array. The option key will be accepted as either `charset` or
-  `encoding`.
 - `doublequote`: Equivalent to the PHP `htmlentities()` native function
   parameter `double_encode`. If set to `false`, existing HTML entities will not
   be encoded. The default is to convert everything (`true`).
-
-  This option must be set via the `$options` parameter or the
-  `setDoubleEncode()` method.
 
 ### Basic Usage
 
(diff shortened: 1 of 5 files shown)


src/HtmlEntities.php
```diff
@@ -4,181 +4,49 @@
...
-use function is_array;
 use function is_scalar;
 use function strlen;
 
 use const ENT_QUOTES;
 
 /**
  * @psalm-type Options = array{
- *     quote_style?: int,
+ *     quotestyle?: int,
```
gsteel, line 17: Whilst I appreciate the desire for consistency, this is a BC break that's not really worth changing - can you revert back to `quote_style` please. This change potentially forces a lot of annoying code changes on users for little benefit.

docs/book/v3/standard-filters.md
```diff
@@ -635,24 +635,17 @@ entity equivalents when possible.
 The following options are supported for `Laminas\Filter\HtmlEntities`:
 
 - `quotestyle`: Equivalent to the PHP `htmlentities()` native function parameter
```
gsteel, line 637: Also `quote_style` here too.

src/HtmlEntities.php
```diff
@@ -199,18 +66,22 @@ public function filter(mixed $value): mixed
         }
         $value = (string) $value;
 
-        $filtered = htmlentities($value, $this->getQuoteStyle(), $this->getEncoding(), $this->getDoubleQuote());
+        $filtered = htmlentities($value, $this->quoteStyle, $this->encoding, $this->doubleQuote);
         if (strlen($value) && ! strlen($filtered)) {
```
gsteel: Can you please tighten up the conditionals here:

```suggestion
        if (strlen($value) > 0 && strlen($filtered) === 0) {
```

src/HtmlEntities.php
```diff
@@ -199,18 +66,22 @@ public function filter(mixed $value): mixed
...
             if (! function_exists('iconv')) {
                 throw new Exception\DomainException('Encoding mismatch has resulted in htmlentities errors');
             }
-            $enc      = $this->getEncoding();
-            $value    = iconv('', $this->getEncoding() . '//IGNORE', $value);
-            $filtered = htmlentities($value, $this->getQuoteStyle(), $enc, $this->getDoubleQuote());
+            $value    = iconv('', $this->encoding . '//IGNORE', $value);
+            $filtered = htmlentities($value, $this->quoteStyle, $this->encoding, $this->doubleQuote);
             if (! strlen($filtered)) {
```
gsteel:
```suggestion
            if (strlen($filtered) === 0) {
```

src/HtmlEntities.php
```diff
@@ -190,7 +58,6 @@ public function setDoubleQuote($doubleQuote)
      * If the value provided is non-scalar, the value will remain unfiltered
      *
      * @throws Exception\DomainException On encoding mismatches.
-     * @psalm-return ($value is scalar ? string : mixed)
      */
     public function filter(mixed $value): mixed
     {
```
gsteel, line 63:
```suggestion
    {
        if (! is_scalar($value) || $value === '') {
```

Checking for an empty string here is defensive against later string length comparisons

test/HtmlEntitiesTest.php
```diff
@@ -44,93 +33,26 @@ public function testBasic(): void
             '"'      => '&quot;',
             '&'      => '&amp;',
```
gsteel: Can we turn this into a data provider please, and add cases for other scalar values that _would_ be filtered, for example `int`, `float`, `bool` - now I think of it, allowing bool here is an odd choice, perhaps change the conditional to `is_string($value) || is_numeric($value)` in `filter()`?

test/HtmlEntitiesTest.php
```diff
@@ -166,32 +88,37 @@ public function testQuoteStyleQuotesEncodeNone(): void
...
     public function testRaisesExceptionIfEncodingMismatchDetectedAndFinalStringIsEmpty(): void
     {
+        $filter = new HtmlEntitiesFilter();
         $string = file_get_contents(__DIR__ . '/_files/latin-1-dash-only.txt');
         $this->expectException(DomainException::class);
-        $this->filter->filter($string);
+        $filter->filter($string);
     }
 
     /** @return list<array{0: mixed}> */
```
gsteel, line 138: This data provider could do with the empty string case, and also boolean

gsteel (changes requested):
Thanks @marcelthole - Sorry it's taken a while to get a review

test/HtmlEntitiesTest.php
gsteel: Also, can you please assert the same output from `__invoke`, otherwise it has no coverage, and explicitly call `$filter->__invoke($foo)` as opposed to `$filter($foo)`

gsteel:
2.40.0 has been released so a rebase will be required

src/HtmlEntities.php
marcelthole: This was only wrong here in the psalm type description. 
As you could see here https://github.com/laminas/laminas-filter/pull/215/files#diff-dddba378a457444b0818056a7af8e629c3a06c8b05e19c2dcc914ff00d304ee0L73 the previous version used already `quotestyle`

```
        if (! isset($options['quotestyle'])) {
            $options['quotestyle'] = ENT_QUOTES;
        }
```

src/HtmlEntities.php
marcelthole: i changed the check also to `is_string` because otherwise a numeric value would be casted to an string

src/HtmlEntities.php
gsteel: Ah yes - That was my fault! 😂

src/HtmlEntities.php
```diff
@@ -206,27 +58,30 @@ public function setDoubleQuote($doubleQuote)
...
-        if (! is_scalar($value)) {
+        if (! is_string($value) || $value === '') {
             return $value;
         }
         $value = (string) $value;
```
gsteel: This cast is now unnecessary

src/HtmlEntities.php
```diff
@@ -206,27 +58,30 @@ public function setDoubleQuote($doubleQuote)
...
 
-        $filtered = htmlentities($value, $this->getQuoteStyle(), $this->getEncoding(), $this->getDoubleQuote());
-        if (strlen($value) && ! strlen($filtered)) {
+        $filtered = htmlentities($value, $this->quoteStyle, $this->encoding, $this->doubleQuote);
+        if (strlen($value) > 0 && strlen($filtered) === 0) {
```
gsteel: The value can no longer have zero length

```suggestion
        if (strlen($filtered) === 0) {
```

gsteel (changes requested):
Thanks @marcelthole - A couple of minor adjustments and this is good to go 👍

src/HtmlEntities.php
marcelthole: Argh so many easy fixes you need to mention :/ for the next PR i will focus a little bit more and also make some code polishing and split the commits to apply the BC changes and one to improve the tests an the code 👍

gsteel (approved):
Thanks @marcelthole 👍

PR #216, opened 2025-01-03
Refactor Dir filter to implement FlterInterface

marcelthole:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | no

### Description
Move Dir filter also to the implementation via FilterInterface and removed the AbstractFilter

### DIFF
--- a/src/Dir.php
+++ b/src/Dir.php
@@ -5,28 +5,29 @@ declare(strict_types=1);
 namespace Laminas\Filter;
 
 use function dirname;
-use function is_scalar;
+use function is_string;
 
 /**
- * @psalm-type Options = array{}
- * @extends AbstractFilter<Options>
+ * @implements FilterInterface<string>
  */
-final class Dir extends AbstractFilter
+final class Dir implements FilterInterface
 {
     /**
      * Defined by Laminas\Filter\FilterInterface
      *
      * Returns dirname($value)
-     *
-     * @psalm-return ($value is scalar ? string : mixed)
      */
     public function filter(mixed $value): mixed
     {
-        if (! is_scalar($value)) {
+        if (! is_string($value) || $value === '') {
             return $value;
         }
-        $value = (string) $value;
 
         return dirname($value);
+    }
+
+    public function __invoke(mixed $value): mixed
+    {
+        return $this->filter($value);
     }
 }
--- a/test/DirTest.php
+++ b/test/DirTest.php
@@ -14,17 +14,24 @@ class DirTest extends TestCase
     /**
      * Ensures that the filter follows expected behavior
      */
-    public function testBasic(): void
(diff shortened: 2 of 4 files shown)


src/Dir.php
```diff
@@ -8,17 +8,14 @@
...
 {
     /**
      * Defined by Laminas\Filter\FilterInterface
      *
      * Returns dirname($value)
-     *
-     * @psalm-return ($value is scalar ? string : mixed)
      */
     public function filter(mixed $value): mixed
     {
```
gsteel, line 21: Can you change this conditional to `if (! is_string($value) && ! is_numeric($value))` and amend the tests to suit please?

I can _nearly_ see a use case for numeric directory names, but casting bool to string and finding the parent directory? Just weird.

gsteel: As this is a refactor, test clean up would be appreciated too:

- Change loops to data providers and include more cases
- Ensure `__invoke` is covered

gsteel (changes requested):
Sorry @marcelthole - reviewed a couple of days ago and never submitted it 🤦‍♂️

src/Dir.php
marcelthole: I updated the testcase. 
For the numeric value i made this filter more strict and only allow strings because i guess that the behavior with numeric values is a little bit strange. https://3v4l.org/5rHL2 

i didn't expect to get the value `"."` as string if i pass `1` as integer value to the filter. If i really want to use that filter for the value i should cast it to a string on my side

gsteel: Yes, `dirname` will return `'.'` for _anything_ in `getcwd`, any non-string input and any non-existent directory/file. It's only remotely useful if it receives something that looks like a fully qualified path, existing on disk or not.

As it stands, the Dir filter is not really useful at all as far as I can tell.

test/DirTest.php
```diff
@@ -14,17 +14,24 @@ class DirTest extends TestCase
...
-            '/path/to/filename'     => '/path/to',
-            '/path/to/filename.ext' => '/path/to',
+        $filter = new DirFilter();
+
+        self::assertSame($expected, $filter($input));
+        self::assertSame($expected, $filter->__invoke($input));
+        self::assertSame($expected, $filter->filter($input));
+    }
+
+    public static function defaultSettingsDataProvider(): array
```
gsteel, line 27: For future reference, please annotate data providers with a precise return types:

```suggestion
    /** @return list<array{0: string, 1: string}> */
    public static function defaultSettingsDataProvider(): array
```

This helps document intent, especially when new items are added or when a larger number of arguments are required.

gsteel (approved):
Thank you @marcelthole :)

PR #222, opened 2025-01-08
Refactor Rename

ramchale:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Refactoring for #177 

I had a couple of questions around the intent for v3;
* I'm assuming removal of getFile and setFile, but wanted to check if we're also removing addFile
* Currently the input for addFile can be a string, or array of `target, etc`, or an array of arrays of `target, etc`. Which of these do we want to support going forward?

### DIFF
--- a/src/File/Rename.php
+++ b/src/File/Rename.php
@@ -4,359 +4,183 @@ declare(strict_types=1);
 
 namespace Laminas\Filter\File;
 
-use Laminas\Filter;
 use Laminas\Filter\Exception;
-use Laminas\Stdlib\ArrayUtils;
-use Traversable;
+use Laminas\Filter\FilterInterface;
 
-use function basename;
-use function count;
+use function array_is_list;
 use function file_exists;
-use function is_array;
-use function is_bool;
+use function fnmatch;
 use function is_dir;
-use function is_scalar;
 use function is_string;
+use function is_writable;
 use function pathinfo;
-use function realpath;
 use function rename;
 use function sprintf;
-use function strlen;
 use function uniqid;
 use function unlink;
 
-use const DIRECTORY_SEPARATOR;
-
 /**
- * @psalm-type Options = array{
- *     file?: array{source?: string, target?: string, overwrite?: bool, randomize?: bool},
- *     ...
+ * @psalm-type OptionsSet = array{
+ *     match?: non-empty-string,
+ *     target_directory?: non-empty-string,
+ *     rename_to?: string,
+ *     overwrite?: bool,
+ *     randomize?: bool
  * }
- * @template TOptions of Options
- * @template-extends Filter\AbstractFilter<TOptions>
+ * @psalm-type Options = OptionsSet|list<OptionsSet>
+ * @psalm-type DefaultedOptionsSet = array{
+ *      match: non-empty-string,
+ *      target_directory: non-empty-string,
+ *      rename_to: string,
+ *      overwrite: bool,
+ *      randomize: bool
+ *  }
+ * @implements FilterInterface<string>
  */
-final class Rename extends Filter\AbstractFilter
+final class Rename implements FilterInterface
 {
-    /**
-     * Internal array of array(source, target, overwrite)
-     *
-     * @var list<array{source: string, target: string, overwrite: bool, randomize: bool}>
-     */
-    protected $files = [];
+    /** @var DefaultedOptionsSet[] */
+    private readonly array $options;
 
-    /**
-     * Options argument may be either a string, a Laminas\Config\Config object, or an array.
-     * If an array or Laminas\Config\Config object, it accepts the following keys:
-     * 'source'    => Source filename or directory which will be renamed
-     * 'target'    => Target filename or directory, the new name of the source file
-     * 'overwrite' => Shall existing files be overwritten ?
-     * 'randomize' => Shall target files have a random postfix attached?
-     *
-     * @param  string|array|Traversable $options Target file or directory to be renamed
-     * @throws Exception\InvalidArgumentException
-     */
-    public function __construct($options = [])
+    /** @param Options $options */
+    public function __construct(array $options = [])
     {
-        if ($options instanceof Traversable) {
-            $options = ArrayUtils::iteratorToArray($options);
-        } elseif (is_string($options)) {
-            $options = ['target' => $options];
-        } elseif (! is_array($options)) {
-            throw new Exception\InvalidArgumentException(
-                'Invalid options argument provided to filter'
-            );
+        $defaultedOptions = [];
+
+        if (array_is_list($options)) {
+            /** @psalm-var OptionsSet $option */
+            foreach ($options as $option) {
+                $defaultedOptions[] = $this->validateAndDefaultOptions($option);
+            }
+        } else {
+            /** @psalm-var OptionsSet $options */
+            $defaultedOptions[] = $this->validateAndDefaultOptions($options);
         }
 
-        $this->setFile($options);
+        $this->options = $defaultedOptions;
     }
 
     /**
-     * Returns the files to rename and their new name and location
-     *
-     * @return list<array{source: string, target: string, overwrite: bool, randomize: bool}>
+     * @param OptionsSet $options
+     * @return DefaultedOptionsSet
      */
-    public function getFile()
+    private function validateAndDefaultOptions(array $options): array
     {
-        return $this->files;
-    }
+        $target = $options['target_directory'] ?? '*';
 
-    /**
-     * Sets a new file or directory as target, deleting existing ones
-     *
-     * Array accepts the following keys:
-     * 'source'    => Source filename or directory which will be renamed
-     * 'target'    => Target filename or directory, the new name of the sourcefile
-     * 'overwrite' => Shall existing files be overwritten?
-     * 'randomize' => Shall target files have a random postfix attached?
-     *
-     * @param  string|array{source?: string, target?: string, overwrite?: bool, randomize?: bool} $options
-     * @return self
-     */
-    public function setFile($options)
-    {
-        $this->files = [];
-        $this->addFile($options);
+        if ($target !== '*') {
+            if (! is_dir($target)) {
+                throw new Exception\InvalidArgumentException(sprintf(
+                    'The target directory "%s" does not exist',
+                    $target
+                ));
+            }
 
-        return $this;
-    }
-
-    /**
-     * Adds a new file or directory as target to the existing ones
-     *
-     * Array accepts the following keys:
-     * 'source'    => Source filename or directory which will be renamed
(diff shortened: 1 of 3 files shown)


gsteel:
I'm not entirely sure because I haven't had a decent look at it myself, but:

- Make use of `Laminas\Filter\File\FileInformation` value object which exists to handle the 3 possible argument types of string|PHP $_FILES|PSRUpload
- I think ideally, expect a single input, i.e. 1 path, or 1 PSR Upload or 1 PHP Upload and return a string representing the new, modified file path _(Or un-filtered input)_
- We may have revise, that, i.e. we might need to accept multiple inputs for PSR/$_FILES but I'd rather avoid that if possible because we then have awkward return type expectations, IMO, successful filtering should === (string) $newFilePath
- If there is a public method other than `filter` or `__invoke`, kill it with 🔥

My reasoning here is that we shouldn't need to worry too much about compat with input filter and form, because they are next in line for major release, and, if you want to deal with multiple uploads, then, from input filters perspective, we should be looking at an 'ArrayInput' wrapping _n_ 'FileInput's rather than all this type juggling.

Options:

- I think `target` should exclusively mean "Target Directory"
- `source` does not make sense and I think it should be removed - the "source" is the filter input
- `overwrite` and `randomize` seem sensible enough

Further reference:

- Existing Filter Docs: https://docs.laminas.dev/laminas-filter/v2/file/#rename
- File uploads with input filter: https://docs.laminas.dev/laminas-inputfilter/file-input/
- File uploads with form: https://docs.laminas.dev/laminas-form/v3/file-upload/

cc @froschdesign @weierophinney

froschdesign:
@gsteel 
> * `source` does not make sense and I think it should be removed - the "source" is the filter input

The `source` option is relevant if several files are to be filtered, see the example in the documentation:

```php
$filter = new \Laminas\Filter\File\Rename([
    [
        'source'    => 'fileA.txt'
        'target'    => '/dest1/newfileA.txt',
        'overwrite' => true,
    ],
    [
        'source'    => 'fileB.txt'
        'target'    => '/dest2/newfileB.txt',
        'randomize' => true,
    ],
]);
```

But I would follow your idea and the `source` option is no longer needed:

> I think ideally, expect a single input, i.e. 1 path, or 1 PSR Upload or 1 PHP Upload and return a string representing the new, modified file path (Or un-filtered input)

I also think that is the right way to go:

> if you want to deal with multiple uploads, then, from input filters perspective, we should be looking at an 'ArrayInput' wrapping _n_ 'FileInput's rather than all this type juggling.

---

If several values are to be filtered, a decorator can still be created for this purpose.

ramchale:
Thanks both. 

Started moving things in that direction. Just a few more questions;
* It seems like 'source' is really only useful for restricting what the filter input can be, which would be fairly key if we're looping through an array of these filters. With that in mind, do we want to rename it to make that more obvious? (Something like 'match_input')
* Currently exceptions are thrown for issues actually renaming the file. This is inconsistent with the other filters, but also this one is slightly unusual in that it modifies things on the server, so double checking that we want to remove these exceptions and return the unfiltered input before I change it?
* Should the file extension be taken from the source file/filter input if not in the target?

gsteel:
OK, so branching configuration based on matching the input to an option set has merit.

Perhaps this can be made more explicit with a `match` option instead of `source` and use `fnmatch` under the bonnet on the basename of the input.

i.e. 

```php
[
    [
        'match' => '*.txt',
        'target' => '/text/files',
    ],
    [
        'match' => '*.pdf',
        'target' => '/pdf/files',
    ],
];
```


We should be receiving a file path (Or PSR Upload, or PHP Upload). All of those 3 evaluate to a single file path input, so the filter's job should be, _(assuming input matches an options set)_:

- Move the file to a target directory with the name unchanged
- Move the file to a target directory and rename it
- Rename the file in-place

In which case, I think that the options to configure that behaviour should be:

- `['target' => '/some/dir', 'renameTo' => '*']`
- `['target' => '/some/dir', 'renameTo' => 'someName.ext']`
- `['target' => '*', 'renameTo' => 'someName.ext']`

I think that `target` should be validated with `is_dir && is_writable` in the constructor _(Unless it's `*`)_.

From there, `randomize`/`overwrite` is trivial, as are the expected outcomes and return value.

The above behaviour makes more sense to me and I can see the filter being useful when processing an HTTP upload, or some other input such as, say, a list of file paths on a local disk.

So, options proposal:

```php
/**
 * @psalm-type OptionsSet = array{
 *   match?: non-empty-string, // default '*'
 *   target?: non-empty-string, // default '*'
 *   renameTo?: non-empty-string, // default '*'
 *   overwrite?: bool, // default false
 *   randomize?: bool, // default false
 * }
 * @psalm-type Options = OptionsSet|list<OptionsSet>
 */
```

- `match` is '*', i.e. match any file name, or, something `fnmatch` can handle
- `target` is '*' i.e. re-name in-place or an existing target directory
- `renameTo` is '*' i.e. don't rename, or, something like `*.txt`, `prefix_*` where * is replaced with `pathinfo()['filename']` or `pathinfo()['basename']` depending on the presence of a `.` in `renameTo`

When match, target and renameTo are all *, then the filter (by default) does absolutely nothing.

This is all just opinion, so feel free to point out glaring stupidity in my suggestions. We have an opportunity to break BC and make it useful and predictable.

> * Currently exceptions are thrown for issues actually renaming the file. This is inconsistent with the other filters, but also this one is slightly unusual in that it modifies things on the server, so double checking that we want to remove these exceptions and return the unfiltered input before I change it?

I think that if the file cannot be successfully moved/renamed, that's an exceptional condition. It means the developer probably hasn't got a writable directory in the right place, so it's more of a configuration error that an exception reacting to user input. If we've been given theoretically filterable input and the filter fails, then it's an exception. For input that cannot be filtered based solely on the input, then return the un-filtered value.

cc @froschdesign

ramchale:
@gsteel  That all makes sense thanks. 

I've put in a single input version, to check the behavior is as expected and ask if there's any further test scenarios anyone can think that need covering. (Ignore the static analysis errors in RenameTest, I just haven't got to them yet).

I've renamed "target" to "target_directory" just to be a bit more explicit.

In terms of passing in an array for configuration for multiple matches, is this something we could do by just adding multiple Rename filters to a field (possibly in a Chain) rather than having an array of arrays for the config? (Happy to add config array handling, but worth checking first)

gsteel:
> In terms of passing in an array for configuration for multiple matches, is this something we could do by just adding multiple Rename filters to a field (possibly in a Chain) rather than having an array of arrays for the config? (Happy to add config array handling, but worth checking first)

I was thinking that the more complex config would be useful for "If it's a word doc, put in /docs, if it's a pdf put in /pdf-files", but you're right that multiple rename filters chained after each other could achieve the same outcome and would also simplify handling configuration internally.

I'm on the fence but leaning towards a single set of options _(And multiple filters if required)_

What do you think @froschdesign?

froschdesign:
Would it then have to be ensured within the chain that different "matches" are defined so that all filters in the chain do not process all files? Or is this simply a misconfiguration?

---

_I think one thing we must not overlook is that this filter is not intended for files that come from an upload._

ramchale:
True it does have the issue that Chains won't stop at the first match. A potential alternative is to have the Rename filter for single cases, and add a RenameList wrapper that can iterate through multiple Rename filters until it finds a match?

marcelthole:
Are there any updates? I also have the PR with the RenameUpload Filter which should handle the Rename stuff for uploaded files. I would like to use the same behavior like here (and with the same naming of the options) if this is done.

ramchale:
> Are there any updates? I also have the PR with the RenameUpload Filter which should handle the Rename stuff for uploaded files. I would like to use the same behavior like here (and with the same naming of the options) if this is done.

Sorry @marcelthole been a busy few weeks. In the absence of further discussion, I'll update this PR to support an array of Option arrays being passed in, as well as just an individual options array

ramchale:
@gsteel @froschdesign a few Psalm issues I need to fix, but otherwise this is functionally what I think it should be doing. Any feedback before I update the docs to reflect the changes in behaviour?

froschdesign:
@gsteel 
This all looks very good to me. Or should we only allow one file and find another solution for multiple files - which is very easy to handle?

froschdesign:
@marcelthole 
>  I would like to use the same behavior like here (and with the same naming of the options) if this is done.

Very good point and this must also be implemented. 👍🏻

ramchale:
If you don't mind having seperate(ish) filters for single and multiple files [something like this](https://github.com/ramchale/laminas-filter/commit/ba1d980495805a46ac95d54791bc32e0021c2d22#diff-1528268bcc2353c7244b56c650abda8e522ae5546b85c6754534430b23e35b49) could work

gsteel:
I haven't forgotten this PR - sorry, very busy!

gsteel (changes requested):
Hey @ramchale Thanks for your patience here 👍 

This looks good to go for me - I like that we can match filenames and move/rename those to specific directories if desired and single file handling is the way to go IMO.

I think your idea of a `RenameList` filter needs further exploration in a separate patch - I'm wondering if `ArrayInput` from `input-filter` covers this need.

If you could fix the minor Psalm issues, we can get this merged I reckon :)

froschdesign:
> …I'm wondering if `ArrayInput` from `input-filter` covers this need.

Good point!

```php
$filter = $this->getFilterChain();

return array_map(
     static fn (mixed $value): mixed => $filter->filter($value),
     $this->value,
);
```
https://github.com/laminas/laminas-inputfilter/blob/47fa98ac6e8b01e07d8ef8bbad2020c2e769e049/src/ArrayInput.php#L35-L47

ramchale:
@gsteel No problem. The code side should be ready for review now. I can add the docs as a follow up (assuming I haven't misinterpreted the conversation)

src/File/Rename.php
```diff
@@ -4,359 +4,183 @@
...
-final class Rename extends Filter\AbstractFilter
+final class Rename implements FilterInterface
 {
-    /**
-     * Internal array of array(source, target, overwrite)
-     *
-     * @var list<array{source: string, target: string, overwrite: bool, randomize: bool}>
-     */
-    protected $files = [];
+    /** @var DefaultedOptionsSet[] */
```
gsteel, line 42: Nitpick: I prefer `/** @var list<DefaultedOptionsSet> */` because it is more explicit and it's harder to miss the trailing `[]`

gsteel (approved):
Thanks very much @ramchale - Sorry the reviews took a long time :)

gsteel:
> I can add the docs as a follow up

This would be very much appreciated 👍 Merging as-is for now

PR #223, opened 2025-01-09
Refactor RenameUpload filter to implement FlterInterface

### LINKED ISSUE #3: PSR-7 file upload parameter naming consistency.
- [x] Is this related to quality assurance?
- [x] Is this related to documentation?
Upgrading my setup to released version of #70 caused some confusion. Updating parameter to help resolve it, and suggest a documentation edit to explain where the instances could come from, as I couldn't figure it out without looking at the diff changes of previous PR refactorings.

Add missing unit tests for new options.

---

Originally posted by @alextech at https://github.com/zendframework/zend-filter/pull/82

marcelthole:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes
| QA            | no

### Description
Move RenameUpload filter also to the implementation via FilterInterface and removed the AbstractFilter


src/File/RenameUpload.php
```diff
@@ -27,207 +28,76 @@
...
+ *     use_upload_extension?: bool,
+ *     overwrite?: bool,
+ *     randomize?: bool,
+ *     stream_factory?: StreamFactoryInterface,
+ *     upload_file_factory?: UploadedFileFactoryInterface,
  * }
- * @template TOptions of Options
- * @extends AbstractFilter<TOptions>
+ * @implements FilterInterface<mixed>
+ * @final
```
marcelthole: Used this `@final` instead of making the class final because otherwise we could not overwrite the `move_uploaded_file` function. 

Or do you prefer to pass a dependency via option that could change the behavior via DependencyInjection? And should the option be documented as well?

src/File/RenameUpload.php
```diff
@@ -237,26 +107,42 @@ public function filter(mixed $value): mixed
...
         // File upload via traditional SAPI
-        if (is_array($value) && isset($value['tmp_name'])) {
-            return $this->filterSapiUploadedFile($value);
+        if (
+            is_array($value)
+            && array_key_exists('tmp_name', $value)
+            && array_key_exists('name', $value)
+        ) {
+            /** @var UploadedFile $uploadedFile */
+            $uploadedFile = $value;
```
marcelthole, line 120: i don't know why psalm will not detect it if i add the `@var` annotation directly to `$value`

src/File/RenameUpload.php
```diff
@@ -237,26 +107,42 @@ public function filter(mixed $value): mixed
...
         return $value;
     }
 
     /**
-     * @param  string $sourceFile Source file path
-     * @param  string $targetFile Target file path
+     * @psalm-suppress MixedReturnStatement
+     * @template T
+     * @param T $value
+     * @return (T is UploadedFileInterface
```
marcelthole, line 136: sadly i had to duplicate this doctype because otherwise psalm will detect the `mixed` return type and mark that as an error for all following calls in the Tests. This comes from the FilterInterface `@return TFilteredValue|T` but i could not define it strict. So i need to use `@implements FilterInterface<mixed>` in this class and overwrite the return type here manually

test/File/RenameUploadMock.php
```diff
@@ -5,36 +5,16 @@
...
- *     use_upload_extension: bool,
- *     overwrite: bool,
- *     randomize: bool,
- *     stream_factory: StreamFactoryInterface|null,
- *     upload_file_factory: UploadedFileFactoryInterface|null,
- *     ...
- * }
- * @template TOptions of Options
- * @extends RenameUpload<TOptions>
+ * @psalm-suppress InvalidExtendClass
```
marcelthole: Expected error because the RenameUpload is marked as `@final`

src/File/RenameUpload.php
marcelthole: I added the deprecated aswell for these methods here #224

gsteel: We should be able to make this final without any problems… and IMO it should be `@implements FilterInterface<string>`, and operate on a single upload at a time. If it allows multiple uploads, then we should change that.

src/File/RenameUpload.php
gsteel: There's a value object you can use in `Filter\File\FileInformation`

```php
if (! FileInformation::isPossibleFile($value)) {
    return $value; // not an upload
}

$file = FileInformation::factory($value);
```

This should also negate the need for the other PSR related deps.

Psalms refusal to accept the forced `@var` is probably because of the templates

gsteel:
@marcelthole I'm not going to review this just yet, because I think that discussion in #222 is relevant here too

src/File/RenameUpload.php
gsteel: I think we need to create a factory for this filter and register it with the "FilterManager". The constructor signature should change to something like:

`__construct(StreamFactoryInterface, UploadedFileFactoryInterface, Closure, Options)`

Where the Closure wraps `move_uploaded_file`, so we can easily inject something else in tests. We can make this closure within the factory instead of registering it as a service.

It _may_ be necessary to make PSR StreamFactory and PSR UploadFactory nullable for compat with MVC.

src/File/MoveUploadedFileInterface.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Filter\File;
+
+use function move_uploaded_file;
+
+final class MoveUploadedFileInterface implements UploadedFileMoverInterface
```
samsonasik: class with `Interface` suffix?

marcelthole: Oh automatic renaming without chercking. Will fix it

- [x] Rename class

gsteel:
Included via #239 

Thanks for getting this one started @marcelthole - I hope I didn't step on your toes too much!

PR #224, opened 2025-01-09
Deprecate getter,setter and protected methods in RenameUpload

marcelthole:
See #223

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1017,6 +1017,27 @@
     <DeprecatedClass>
       <code><![CDATA[AbstractFilter]]></code>
     </DeprecatedClass>
+    <DeprecatedMethod>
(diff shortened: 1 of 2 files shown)


psalm-baseline.xml
```diff
@@ -836,6 +840,9 @@
       <code><![CDATA[Filter\Decrypt]]></code>
       <code><![CDATA[parent::filter($content)]]></code>
     </DeprecatedClass>
+    <DeprecatedInterface>
```
marcelthole: unsure why theses entries are added. I only added the deprecations and run `psalm --set-baseline=psalm-baseline.xml` with the psalm version of the lock file

marcelthole: `--no-cache` did the trick

gsteel (approved):
Thank you @marcelthole 👍

PR #225, opened 2025-01-26
Drop Support for PHP 8.1, Upgrade Psalm to v6 & PHPUnit to v11

gsteel:
- Marks _(nearly)_ all filter classes as readonly
- Fixes a number of new Psalm issues, mainly concerned with null/false return types from built-ins

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
             "composer/package-versions-deprecated": true
         },
         "platform": {
-            "php": "8.1.99"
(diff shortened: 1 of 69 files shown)


gsteel:
Needs https://github.com/laminas/laminas-servicemanager/pull/255

gsteel:
Ugh - closing - we're not dropping 8.1 so filters cannot be `final readonly`

PR #228, opened 2025-02-03
Upgrade Psalm to v6

gsteel:
Fixes a number of new Psalm issues, mainly concerned with null/false return types from built-ins


Ocramius (approved):
🚢

PR #232, opened 2025-03-03
Introduce Composer Require Checker along with a Makefile

gsteel:
- Add CRC as a standalone tool
- Add CI job for CRC
- Add Makefile for running local QA checks - the CI job also uses make for CRC
- Add used deps and extensions to composer.json
- Add soft required extensions to require-dev, laminas-ci config and CRC ignores


Ocramius (approved):
Thanks @gsteel!

PR #234, opened 2025-03-24
Improve deprecation messages in `FilterChain`

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### DIFF
--- a/src/FilterChain.php
+++ b/src/FilterChain.php
@@ -70,7 +70,8 @@ class FilterChain extends AbstractFilter implements Countable, IteratorAggregate
     }
 
     /**
-     * @deprecated This method will be removed in 3.0.0 without replacement
+     * @deprecated This method will be removed in 3.0.0 without replacement. Future versions of FilterChain will require
+     *             that all options are provided at construction time.
      *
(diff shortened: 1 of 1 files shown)


src/FilterChain.php
```diff
@@ -70,7 +70,8 @@ public function __construct($options = null)
     }
 
     /**
-     * @deprecated This method will be removed in 3.0.0 without replacement
+     * @deprecated This method will be removed in 3.0.0 without replacement. Future versions of FilterChain will require
```
froschdesign, line 73: An alternative can be a link to the migration guide.

froschdesign (approved):

src/FilterChain.php
akrabat: Links in addition to the message would be good when we have a link.

akrabat (approved):

src/FilterChain.php
gsteel: The "preparing-for" docs are out of date right now, and the migration guide un-released so I've merged as-is for now

PR #239, opened 2025-04-12
Refactor `File\RenameUpload`

gsteel:
Includes changes in #223

This patch further simplifies the filter, making use of `FileInformation` to handle the different possible acceptable inputs, and helps get rid of the awkward additional PSR dependencies by only returning the file path of the uploaded file.

Adds the `MoveUploadedFile` utility as a constructor dependency, requiring a factory setup with the plugin manager.

Expands `MoveUploadedFile` to take care of basic error handling

Reverts `target_directory` option and semantics back to `target`, hopefully, for a simpler transition.

I'd appreciate review and input from @ramchale, @marcelthole and @froschdesign 👍


src/File/RenameUploadFactory.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    public function __invoke(
+        ContainerInterface $container,
+        string $requestedName,
+        array|null $options = null,
+    ): RenameUpload {
+        /**
+         * @psalm-suppress MixedArgumentTypeCoercion Runtime validation of options is not worth it.
+         */
+        return new RenameUpload($options ?? [], new MoveUploadedFile());
```
froschdesign: Have I overlooked something or why is the factory needed?

gsteel: Arguably, we could inline the `MoveUploadedFile` class. It was extracted so that we could stub PHP's `move_uploaded_file` function, but now the default implementation checks `is_uploaded_file` _(and runs `rename` if its not)_, it feels less important that its a stand-alone collaborator. The only other benefit of keeping it external is that it now also runs an error handler to upcast PHP errors/warnings to exceptions, but we could also just `new` it in the constructor and drop the factory completely.

What do you think?

src/File/MoveUploadedFile.php
```diff
@@ -0,0 +1,47 @@
...
+        set_error_handler(static function (
+            int $errorNumber,
+            string $message,
+            $file = null,
+            $line = null,
+        ): never {
+            throw new ErrorException($message, $errorNumber, $errorNumber, $file, $line);
+        });
+
+        if (! is_uploaded_file($sourceFile)) {
```
marcelthole: Isn't this the security feature of `move_uploaded_file` that will be disabled with this condition?
This Filter should be used for uploaded files only. If someone hijack the $_FILES object this security mechanism will not work anymore. 

You need this only for the Tests and there is already the MoveUploadedFileInterface. Would prefer to use here the move_uploaded_file method and use the rename Handler only for the tests

gsteel: Good point, running `is_uploaded_file` in the filter and rejecting input upfront would cover this. We could then continue to use the existing (And internal) `MoveUploadedFile` implementation in the knowledge that the conditional there is only used in test scenarios.

src/File/RenameUploadFactory.php
froschdesign: > …but we could also just `new` it in the constructor and drop the factory completely.

It is already there:

```php
$this->moveUploadedFile = $moveUploadedFile ?? new MoveUploadedFile();
```

https://github.com/laminas/laminas-filter/pull/239/files#diff-94631e5208b9361631c5713c4af740513be7608554440d9130f6e8f633498bd5R62

gsteel: Thanks @froschdesign - I've dropped the factory now 👍

src/File/MoveUploadedFile.php
gsteel: Brains aren't working well today - thanks for the review @marcelthole - I've reverted closer to your original work 👍

gsteel:
If we can get this over the line before the end of the month, we can vote on releasing v3 at the next TSC meet in May.

ramchale (approved):
Looks like a good simplification to me

docs/book/v3/file.md
```diff
@@ -166,6 +166,8 @@ echo $filter->filter('fileB.txt');
 
 `Laminas\Filter\File\RenameUpload` can be used to rename or move an uploaded file to a new path.
 
+It will accept an array representing a single [PHP file upload](https://www.php.net/manual/features.file-upload.post-method.php), a PSR7 [`UploadedFileInterface`](https://www.php-fig.org/psr/psr-7/#36-psrhttpmessageuploadedfileinterface), or a string representing the file path of an uploaded file.
```
froschdesign: An unordered list for the types would give a better overview.

docs/book/v3/file.md
```diff
@@ -203,13 +197,11 @@ The following set of options are supported:
...
 
-`RenameUpload` does not support an array of options like the`Rename` filter.
-When filtering HTML5 file uploads with the `multiple` attribute set, all files
-will be filtered with the same option settings.
+`RenameUpload` can only operate on one file at a time.
 
 ### Usage Examples
 
-Move all filtered files to a different directory:
+Move all filtered files to a different directory in a Laminas MVC application:
```
froschdesign: I do not understand the addition. Does not this also work in simple scripts without laminas-mvc?

gsteel: I've removed that addition and changed one of the examples to use `$_FILES` directly.

Also, dropped the second, nullable arg to the constructor in the examples which is unnecessary and probably will just cause confusion for users

froschdesign: Thank you! 👍🏻 

In principle, the components should always be described as they function on their own - if this is possible, of course.
The explanations for Mezzio, laminas-mvc or Slim & Co. must be described separately under a separate navigation point / page.

marcelthole (approved):

froschdesign (approved):

PR #240, opened 2025-04-12
Add `ToEnum` filter

gsteel:
This is a re-implementation of #94 which seems to be abandoned.

### DIFF
--- a/src/ToEnum.php
+++ b/src/ToEnum.php
@@ -1,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Filter;
+
+use BackedEnum;
+use UnitEnum;
+
+use function assert;
+use function constant;
+use function defined;
+use function is_a;
+use function is_int;
+use function is_string;
+
+/**
+ * @psalm-type Options = array{
+ *     enum: class-string<UnitEnum>|class-string<BackedEnum>,
+ * }
+ * @implements FilterInterface<UnitEnum|BackedEnum>
+ */
+final class ToEnum implements FilterInterface
+{
+    /** @var class-string<UnitEnum>|class-string<BackedEnum> */
+    private readonly string $enum;
+
+    /** @param Options $options */
+    public function __construct(array $options)
+    {
+        $this->enum = $options['enum'];
+    }
+
+    public function filter(mixed $value): mixed
+    {
+        if (! is_int($value) && ! is_string($value)) {
+            return $value;
+        }
+
+        if (is_a($this->enum, BackedEnum::class, true)) {
+            $enum         = null;
+            $stringBacked = is_string($this->enum::cases()[0]->value);
+            if ($stringBacked && is_string($value)) {
+                $enum = $this->enum::tryFrom($value);
+            }
+
+            if (! $stringBacked && (is_int($value) || ((string) (int) $value === $value))) {
+                $enum = $this->enum::tryFrom((int) $value);
+            }
+
+            if ($enum !== null) {
+                return $enum;
+            }
+        }
+
+        if (! is_string($value)) {
+            return $value;
+        }
+
+        $constantName = $this->enum . '::' . $value;
+
+        if (defined($constantName)) {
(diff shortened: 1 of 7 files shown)


src/ToEnum.php
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+
+        if (! is_string($value)) {
+            return $value;
+        }
+
+        $constantName = $this->enum . '::' . $value;
+
+        if (defined($constantName)) {
```
froschdesign, line 62: Where is this part tested and how does this correspond with the statement in the documentation?

> This filter requires a single option `enum` which must be a class-string representing a PHP enum.

src/ToEnum.php
```diff
@@ -0,0 +1,77 @@
...
+            if (! $stringBacked && (is_int($value) || ((string) (int) $value === $value))) {
+                $enum = $this->enum::tryFrom((int) $value);
+            }
+
+            if ($enum !== null) {
+                return $enum;
+            }
+        }
+
+        if (! is_string($value)) {
```
froschdesign, line 56: Seems like a repetition to me.

froschdesign (changes requested):

src/ToEnum.php
gsteel: It's not, at this point `$value` is `int|string` and `int` is no longer acceptable because the value _must_ correspond to a case **name** to be filtered to an enum

src/ToEnum.php
gsteel: At this point, `$value` is a string that will correspond to `SomeEnum::$value` (or not). If the string is a valid case name, `defined` will return true.

I'm not quite sure what you're getting at here.

froschdesign: You are right, somehow I completely misinterpreted the property. 🤦🏻‍♂️

src/ToEnum.php
```diff
@@ -0,0 +1,77 @@
...
+ */
+final class ToEnum implements FilterInterface
+{
+    /** @var class-string<UnitEnum>|class-string<BackedEnum> */
+    private readonly string $enum;
+
+    /** @param Options $options */
+    public function __construct(array $options)
+    {
+        $this->enum = $options['enum'];
```
froschdesign, line 31: No check for the array key?

gsteel: Theoretically static analysis has our back here because the `Options` type is `array{enum: ...}`.

I can change this to be ultra defensive and validate the option is set, and, `class-string<UnitEnum>|class-string<BackedEnum>`.

froschdesign: My problem is much simpler:

```php
$filter = new Laminas\Filter\ToEnum([]);
```

How does this end or what does it lead to?

gsteel: This is a `TypeError` assigning null to typed property string, so a hard crash.

gsteel: Preceded by a warning also for the undefined array key

froschdesign: I am sorry, I ca not remember, but is that what we want? Or rather throw an exception?
It seems that is what we want:

https://github.com/laminas/laminas-filter/blob/3d0607220564380a6276eda3078f64948760e8f1/src/DecompressArchive.php#L34-L39

I am thinking about how we should document this.

froschdesign (approved):

src/ToEnum.php
gsteel: Very good point.

In this case, if we add defensive checks, Psalm will moan at us about the redundant checks with good reason, because the required type is precisely defined.

In the case you quoted above, Psalm cannot tell the user that an arbitrary string is not a directory, so the defensive check is required.

Overall, the defensive checks provide a better user experience for those not using static analysis by way of clear exception messages and traces.

In the case of filter, SA is a bit of a moot point because inference will not occur when filters are thrown together from an array input-filter specification.

Not sure what the best approach is.

Maybe @Ocramius or @weierophinney could chime in with opinions on best-practice here.

PR #241, opened 2025-04-14
Improve the `ToString` filter to cast array members to strings recursively

marcelthole:
According to https://github.com/laminas/laminas-filter/issues/177 this should be the last Task:
- `ToString` needs to apply recursively to arrays to be more useful + docs need updating

### DIFF
--- a/docs/book/v3/migration/v2-to-v3.md
+++ b/docs/book/v3/migration/v2-to-v3.md
@@ -365,6 +365,12 @@ The following methods have been removed:
 
 The constructor now only accepts an associative array of [documented options](../standard-filters.md#tonull).
 
+#### `ToString`
+
+This filter now also casts array values to strings recursively.
+
+[Basic Usage](../standard-filters.md#tostring)
+
 #### `UnderscoreToSeparator`
 
 The constructor now only accepts an associative array of [documented options](../word.md#underscoretoseparator).
--- a/docs/book/v3/standard-filters.md
+++ b/docs/book/v3/standard-filters.md
@@ -862,6 +862,8 @@ It is best practice is to use the `TYPE_*` constants rather than the human-reada
 ## ToString
 
 The `ToString` filter casts `Stringable` objects or scalar values to `string`.
+If an array is received, member values are cast to a string recursively.
+
 This filter has no runtime options.
 
 ### Basic Usage
@@ -870,14 +872,16 @@ This filter has no runtime options.
 $filter = new \Laminas\Filter\ToString();
 
 $filter->filter(123); // "123"
+$filter->filter(['id' => 100, 'roles' => null]); // ['id' => "100", 'roles' => null]
 ```
 
-Non-scalar input will be returned un-filtered:
+Non-scalar or null input will be returned un-filtered:
 
 ```php
 $filter = new \Laminas\Filter\ToString();
 
-$filter->filter(['muppet' => 'Kermit']); // ['muppet' => 'Kermit']
+$filter->filter(null); // null
+$filter->filter(new \stdClass()); // \stdClass object
 ```
 
 ## PregReplace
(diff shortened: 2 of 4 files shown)


docs/book/v3/standard-filters.md
```diff
@@ -861,7 +861,7 @@ It is best practice is to use the `TYPE_*` constants rather than the human-reada
 
 ## ToString
 
-The `ToString` filter casts `Stringable` objects or scalar values to `string`.
+The `ToString` filter casts `Stringable` objects or scalar values to `string`. Iterable values will be casted recursive
```
marcelthole: Unsure if it make sense to make this configurable with an option if the user want to use it or not

src/ToString.php
```diff
@@ -18,6 +19,15 @@ final class ToString implements FilterInterface
...
     {
+        if (is_iterable($value)) {
+            /** @psalm-suppress MixedAssignment */
+            foreach ($value as &$item) {
+                /** @psalm-suppress MixedAssignment */
+                $item = $this->filter($item);
+            }
+            unset($item);
+            return $value;
+        }
```
gsteel: Could you take a look at `ScalarOrArrayFilterCallback::applyRecursively` and use that instead?

docs/book/v3/standard-filters.md
gsteel: I don't think an option is necessary as long as the documented behaviour is clear

src/ToString.php
marcelthole: Ah didn't know that. Sure is changed :)

Ah psalm problem again :/ 
`ERROR: MixedArgumentTypeCoercion - src/ToString.php:25:17 - Argument 2 of Laminas\Filter\ScalarOrArrayFilterCallback::applyRecursively expects Closure(string):string, but parent type impure-Closure(mixed):mixed provided (see https://psalm.dev/194)`

is it fine to change the signature for the closure to `@param Closure(mixed): mixed $callback` ? 
or at least null|iterable|object|resource (if i didn't missed a type)

docs/book/v3/standard-filters.md
```suggestion
The `ToString` filter casts `Stringable` objects or scalar values to `string`.
If an array is received, member values are cast to a string recursively.
```

src/ToString.php
```diff
@@ -18,6 +19,12 @@ final class ToString implements FilterInterface
      */
     public function filter(mixed $value): mixed
     {
+        if (is_iterable($value)) {
+            return ScalarOrArrayFilterCallback::applyRecursively(
+                $value,
+                $this->filter(...)
+            );
+        }
```
gsteel:
```suggestion
        return ScalarOrArrayFilterCallback::applyRecursively(
            $value,
            static fn (string $value): string => $value,
        );
```

Casting is applied inside `applyRecursively`, so this filter literaly has nothing to do.

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -365,6 +365,12 @@ The following methods have been removed:
 
 The constructor now only accepts an associative array of [documented options](../standard-filters.md#tonull).
 
+#### `ToString`
+
+This filter casts now also iterable values recursively to strings.
```
gsteel:
```suggestion
This filter now also casts array values to strings recursively.
```

gsteel (changes requested):

src/ToString.php
marcelthole: ah dumb me. The ScalarOrArrayFilterCallback will do the string cast already. So could replace the filter logic completly.

gsteel (approved):
Thanks @marcelthole 💪

Nearly there - I think there's still some work on the docs todo, but this is likely the last code changes for v3 🎉

PR #254, opened 2025-06-12
Add documentation for the Filter Plugin Manager

froschdesign:
Introduce a new section detailing the Filter Plugin Manager in `v3` documentation. This covers its purpose, usage examples, custom filter registration, and relevance in input filters and forms, enhancing user understanding and accessibility.


docs/book/v3/plugin-manager.md
```diff
@@ -0,0 +1,205 @@
...
+$filter = $filterPluginManager->get('StringTrim');
+```
+
+TIP: The filter plugin manager is not really necessary for all filters which are provided by laminas-filter, _[beside the filter chains](filter-chains.md)_.
+Because all filters can be created without external dependencies, directly by using the constructor.
+(The filter chains require the filter plugin manager as a dependency to create the filters they contain.)
+
+## Registering Custom Filters
+
+Custom filters can be registered with the filter plugin manager using the `configure` method or bypassing configuration to the constructor.
```
gsteel:
```suggestion
Custom filters can be registered with the filter plugin manager using the `configure` method or by passing configuration to the constructor.
```

docs/book/v3/plugin-manager.md
```diff
@@ -0,0 +1,205 @@
...
+}
+```
+
+The filter plugin manager can create the custom filter by the related class name:
+
+```php
+$filter = $filterPluginManager->get(ExampleFilter::class);
+```
+
+The manager uses [the factory `Laminas\ServiceManager\Factory\InvokableFactory`](https://docs.laminas.dev/laminas-servicemanager/v4/configuring-the-service-manager/#factories) to instantiate the filter, and will also pass the options for the filter to the constructor:
```
gsteel: Is it worth documenting the `build` method, which is the only way for client code to pass options to the filter plugin manager for filter constructors?

docs/book/v3/plugin-manager.md
```diff
@@ -0,0 +1,205 @@
...
+
+```php
+$filter = $filterPluginManager->get(
+    ExampleFilter::class,
+    [
+        // Options for the filter
+    ]
+);
```
gsteel: Ah yes - this is not correct. `get` no longer has `$options`

https://github.com/laminas/laminas-servicemanager/blob/04f919d9dafb1d19436ed289673d1a867b772fe0/src/AbstractPluginManager.php#L187

docs/book/v3/plugin-manager.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+WARNING: An alias for the custom filter is not automatically created.
+If an alias is to be used, [it must be registered manually](#registering-custom-filters) in the filter plugin manager configuration.
+
+## Why Is the Filter Plugin Manager Relevant?
+
+The filter plugin manager is relevant because it is used in [input filters of laminas-inputfilter](https://docs.laminas.dev/laminas-inputfilter/), stand-alone or in [forms of laminas-form](https://docs.laminas.dev/laminas-form/).
```
gsteel: Just a thought, that a simple sentence to describe its main purpose might be:

> The primary purpose of the `FilterPluginManager` is to provide a single object that can retrieve _any_ known filter, regardless of whether it has complex dependencies or not.

docs/book/v3/plugin-manager.md
```diff
@@ -0,0 +1,205 @@
...
+Internally, the manager is used to create filters for the input filter.
+
+The following form illustrates how the filters are defined for the input filter.
+The filters are automatically fetched from the filter plugin manager by their class name or alias.
```
gsteel:
```suggestion
The filters are automatically retrieved from the filter plugin manager by their class name or alias.
```

gsteel: Is it worth noting that filters are not "shared", i.e.

```php
$plugins->get(SomeFilter::class) !== $plugins->get(SomeFilter::class);
```

docs/book/v3/plugin-manager.md
froschdesign: You are right!

gsteel (approved):
Thanks @froschdesign 👍

PR #258, opened 2025-08-08
Allow filter chains to be constructed with instances in the specification

gsteel:
We should be able to create a filter chain with something like:

```php
$chain = new FilterChain($pluginManager, [
    'filters' => [
        [ /** Spec Array */ ],
        new StringTrim(),
    ],
]);
```

This patch allows instances to given in the spec array and updates the documented types to suit

Relevant to:

https://github.com/laminas/laminas-inputfilter/pull/149

### DIFF
--- a/src/FilterChain.php
+++ b/src/FilterChain.php
@@ -11,6 +11,7 @@ use Psr\Container\ContainerExceptionInterface;
 use Traversable;
 
 use function count;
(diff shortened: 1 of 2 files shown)


Ocramius (approved):

Ocramius:
Thanks @gsteel!

PR #259, opened 2025-08-11
Bump support for laminas-servicemanager to 4.4.0

settermjd:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

While not strictly necessary for _laminas-filter_, I've created this change to support the latest version of _laminas-servicemanager_ that contains [the removes duplicate delegators fix](https://github.com/laminas/laminas-servicemanager/pull/251). By including this change, I can then update _laminas-inputfilter_ to support the same version of laminas-servicemanager as well, as it depends on laminas-filter.

As some background, the reason for the version bump is that a project I'm working on depends on _laminas-servicemanager_ and _laminas-inputfilter_, but needs to have the above fix, as without it when loaded into a Mezzio project, its ConfigProvider classes, which loads a set of custom routes, will cause a duplicate route exception if the existing project's ConfigProvider class also loads routes into the applications routing table.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         "ext-filter": "*",
         "ext-iconv": "*",
         "ext-mbstring": "*",
-        "laminas/laminas-servicemanager": "^4.1.0",
+        "laminas/laminas-servicemanager": "^4.4.0",
         "laminas/laminas-stdlib": "^3.19.0",
         "psr/container": "^1 || ^2",
         "psr/http-factory": "^1.1",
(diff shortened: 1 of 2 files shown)


gsteel:
@settermjd It'll be a while before input-filter is released with SMv4 support as it's a pretty big change.

It will likely support `^4.x` and so it will be unnecessary to bump to a specific version - any project running `composer update` will get latest deps and all the mentioned fixes.

To rephrase, I don't see the point - it's not the role of `filter` or `input-filter` to fix mezzio's / other projects deps, particularly when the current dependency set would not prevent an update to `4.whatever` unless I'm missing something?

PR #261, opened 2025-08-13
Fix filters registered under `services` cannot be found (no factory)

gsteel:
Because `get()` no longer supports options (to be compliant with psr-container), filter chain used `build()` to retrieve all filter instances.

This is problematic when the user has configured a filter instance under `services`, because `build()` of-course does not fetch from `services`.

The fix here is to call `get()` when no options are given.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,13 +46,13 @@
         "ext-bz2": "*",
         "ext-zlib": "*",
         "ext-zip": "*",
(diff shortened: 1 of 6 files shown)


gsteel:
This is relevant to https://github.com/laminas/laminas-inputfilter/pull/149

froschdesign (approved):

PR #263, opened 2025-10-13
Drop support for PHP 8.1, Add support for PHP 8.5

gsteel:
Adding support for 8.5 to the 2.x series because installation is not yet widespread for v3

### DIFF
--- a/src/FilterChain.php
+++ b/src/FilterChain.php
@@ -295,6 +295,17 @@ class FilterChain extends AbstractFilter implements Countable, IteratorAggregate
         return ['filters'];
     }
 
+    /**
+     * @internal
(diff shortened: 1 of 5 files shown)


src/FilterChain.php
```diff
@@ -295,6 +295,17 @@ public function __sleep()
...
+     * @internal
+     * @deprecated This method will be removed in 3.0.0 without replacement
+     *
+     * @see __sleep()
+     */
+    public function __serialize(): array
+    {
+        return ['filters' => $this->filters];
+    }
+
```
gsteel, line 308: Filter chains are not serializable in v3 so this can be dropped during mergeup

Ocramius (approved):

Ocramius:
Thanks @gsteel!

